PR #26, opened 2019-04-04
Task structure

fynnfeldpausch:
* updated the task structure
* implemented a more modular design
* improved UI/UX


src/main/webapp/app/tasks/task.service.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (environment.mockBackend) {
+      return of(this.readTasks());
+    }
+    return of([]); // TODO: Implement http call
+  }
+
+  /**
+   * Adds a new task to the list of tasks.
+   *
+   * @param task the task's name
```
IamBatman222: param should be "name"

src/main/webapp/app/app.component.ts
```diff
@@ -1,5 +1,9 @@
-import {Component, OnInit} from '@angular/core';
```
IamBatman222: You forgot to adapt the unit test

fynnfeldpausch:
I quite like the overflow, to be honest :-D

IamBatman222:
> I quite like the overflow, to be honest :-D

Yes, this one is just a matter of taste. I can also live with the overlapping =)

src/main/webapp/app/app.component.ts
fynnfeldpausch: Added all unit tests except for the task service. We should wait until the HTTP parts are implemented.

PR #48, opened 2019-09-01
Set Due Date for a Task

emrep:
- the user can set a due date for a task
- the task list is sorted by due date
- the user can also set a due time for a task


fynnfeldpausch:
Hi Emre,

thanks for your PR. It looks good and I just have a few minor remarks and ideas for it:
* if you add a date picker to select a date, why not also use a time picker to select the time
* the time input label is always truncated due to its length
* the date picker icon does not have the correct text color
* tests for the date time localizations are missing

Cheers, Fynn

PR #63, opened 2019-10-07
Adding functionality to set a task as done and sort tasks based on status [Frontend Part]

cwiemeyer:
Basic functionality added to set a task as done via material icon button and by using a Boolean property to flag done tasks. Showing visual feedback when setting a task as done.

Also tasks set as done will be sorted below open tasks.

Jasmine/Karma tests adjusted and added as well.
Currently works with the local version of the application (by using `yarn start-local`).
Flag and basic functionality for the backend is prepared, but needs further work on it. Suggestions to solve in a clean way for the backend are welcome.

If you like my work so far, it would be a pleasure for me to add the backend part and to add further tests as well in a next step.

Thanks for your feedback in advance and kind regards.


src/main/resources/db/migration/V1__Initial_Setup.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE task (
     id VARCHAR(36) CONSTRAINT task_id_pkey PRIMARY KEY,
     name VARCHAR (128) NOT NULL,
-    created TIMESTAMP WITH TIME ZONE NOT NULL
+    created TIMESTAMP WITH TIME ZONE NOT NULL,
```
fynnfeldpausch: Why did you include backend migrations in the frontend task?

src/main/webapp/app/tasks/default-task.service.spec.ts
```diff
@@ -61,4 +61,16 @@ describe('DefaultTaskService', () => {
...
     req.flush({});
   });
+
+  it('should update a task', () => {
+    // when
+    taskService.update({id: 'id123', name: 'My task', isDone: true}).subscribe();
+
+    // then
+    const req = httpTestingController.expectOne(request => request.url === 'http://backend.tld/tasks/id123');
+    expect(req.request.method).toEqual('PUT');
```
fynnfeldpausch: I would expect the payload to be tested as well.

fynnfeldpausch: I would expect the payload to be tested as well.

src/main/webapp/app/tasks/default-task.service.ts
```diff
@@ -20,6 +20,10 @@ export class DefaultTaskService implements TaskService {
     return this.http.delete<void>(this.baseUrl + '/tasks/' + id);
   }
 
+  update(task: Task): Observable<void> {
```
fynnfeldpausch: I would expect the `update` function to return the updated task.

fynnfeldpausch: Also, one could argue that the `update` function should take two parameters: `update(id: string, updates: Partial<Task>)`

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,12 +32,32 @@ export class LocalTaskService implements TaskService {
...
 
+  update(task: Task): Observable<void> {
+    const tasks = this.readTasks();
+    const index = tasks.findIndex(taskItem => taskItem.id === task.id);
+    if (index !== -1) {
+      tasks.splice(index, 1, task);
+      this.writeTasks(tasks);
+      this.getAll();
+    }
+    return of(null);
```
fynnfeldpausch: From an API design viewpoint it could make sense to return the edited task as an `Observable< Task>`

Issue #29, opened 2019-04-05
A user can mark a task as done


fynnfeldpausch:
As a user, I want to **mark tasks as done** so that I can see which tasks have already been worked on.

**Acceptance criteria**
- the user can mark a task as done
- the task list is sorted by done status
  - tasks that are not done go to the top of the list
  - tasks that are done go to the bottom of the list

**Optional acceptance criteria**
- the user clear the list of done tasks

**Additional information**
None.

PR #64, opened 2019-10-08
Enable task marked as done

NestorSegura:
work in progress to enable task be marked as done #29


src/main/java/com/coyoapp/tinytask/web/TaskController.java
```diff
@@ -44,4 +38,11 @@ public void deleteTask(@PathVariable String taskId) {
     log.debug("deleteTask(taskId={})", taskId);
     taskService.deleteTask(taskId);
   }
+
+  @ResponseStatus(HttpStatus.OK)
+  @PostMapping("/marktaskdone/{taskId}")
```
fynnfeldpausch: There are a few things that come to mind here:
- as this is basically an update of a given task, PUT would be more appropriate instead of POST
- following API guidelines, the {taskId} should be at the start of the mapping
- naming could also be simplified, as the "task" is already contained in the mapping definition of the controller - one could go for `@PutMapping("/{taskId}/markdone")` or similar

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,6 +32,18 @@ export class LocalTaskService implements TaskService {
     return of(null);
   }
 
+  markAsDone(id: string): Observable<Task> {
+    const tasks = this.readTasks();
+    const index = tasks.findIndex(task => task.id === id);
+    let newTask = tasks[index];
```
fynnfeldpausch: What happens here if `index === -1`?

src/main/resources/db/migration/V3__fix_value_type.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE task
```
fynnfeldpausch: Why are you recreating the table column with exactly the same constraints?

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,10 +1,13 @@
-<mat-list data-cy="task-list">
+<mat-list [ngStyle]="{'pointer-event': markedDone ? 'none' : 'auto'}" data-cy="task-list">
```
fynnfeldpausch: This is not a good way to give feedback to the user.

src/main/webapp/app/tasks/default-task.service.ts
```diff
@@ -23,4 +23,8 @@ export class DefaultTaskService implements TaskService {
   getAll(): Observable<Task[]> {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
   }
+
+  markAsDone(id: string): Observable<Task> {
```
fynnfeldpausch: This functionality should be covered by tests.

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,6 +32,18 @@ export class LocalTaskService implements TaskService {
     return of(null);
   }
 
+  markAsDone(id: string): Observable<Task> {
```
fynnfeldpausch: This functionality should be covered by tests.

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -16,13 +16,24 @@ export class TaskListComponent {
...
+
   constructor(@Inject('TaskService') private taskService: TaskService) { }
 
   delete(task: Task): void {
     this.taskService.delete(task.id).subscribe(() => {
       this.deleted.emit(task);
     });
   }
+
+  markAsDone(task: Task): void {
```
fynnfeldpausch: This functionality should be covered by tests.

PR #66, opened 2019-10-29
Search task

Ratatosk32:
BE part: basic query for getting tasks by name


src/main/java/com/coyoapp/tinytask/service/DefaultTaskService.java
```diff
@@ -31,10 +33,12 @@ public TaskResponse createTask(TaskRequest taskRequest) {
   }
 
   @Override
-  @Transactional(readOnly = true)
-  public List<TaskResponse> getTasks() {
-    log.debug("getTasks()");
-    return taskRepository.findAll().stream().map(this::transformToResponse).collect(toList());
+  public List<TaskResponse> getTasks(String query) {
+    return taskRepository
+      .findAll(getNameFilter(query))
```
TimSielemann: You could create a method inside the repository named `findByNameContaining(String name)` and get rid of the filter implementation here because spring is creating that then out of the box.

PR #71, opened 2019-12-08
A user can mark a task as done

jullierme:
As a user, I want to mark tasks as done so that I can see which tasks have already been worked on.

- [x] the user can mark a task as done
- [x] the task list is sorted by done status
    - [x] tasks that are not done go to the top of the list
    - [x] tasks that are done go to the bottom of the list
- [x] the user can clear the list of done tasks


jullierme:
Sorry, just realize that I forgot to run **ng lint**. How about add [husky](https://github.com/typicode/) to run **ng lint** before commit the code.

Issue #28, opened 2019-04-05
A user can search for tasks


fynnfeldpausch:
As a user, I want to **search all tasks** so that I can find relevant tasks quickly.

**Acceptance criteria**
- the user can search tasks via a search input
- the search is performed instantaneously

**Optional acceptance criteria**
- the user can deep-link to a search result
- the search can be reset with a single click

**Additional information**
To avoid superfluous requests the search input should be throttled.

PR #72, opened 2019-12-20
"A user can search for tasks"

koddist:
Simple search input for tasks #28

### DIFF
--- a/src/main/webapp/app/app.component.ts
+++ b/src/main/webapp/app/app.component.ts
@@ -26,6 +26,10 @@ export class AppComponent implements OnInit {
     this.tasks$ = this.taskService.getAll();
   }
 
+  search(query): void {
+    this.tasks$ = this.taskService.getAll(query);
+  }
+
   created(): void {
     this.tasks$ = this.taskService.getAll();
   }
--- a/src/main/webapp/app/tasks/local-task.service.ts
+++ b/src/main/webapp/app/tasks/local-task.service.ts
@@ -10,8 +10,8 @@ export class LocalTaskService implements TaskService {
 
   private static readonly STORAGE_KEY: string = 'tiny.tasks';
 
-  getAll(): Observable<Task[]> {
-    return of(this.readTasks());
+  getAll(query?:string): Observable<Task[]> {
+    return of(this.readTasks(query));
   }
 
   create(name: string): Observable<Task> {
@@ -32,9 +32,22 @@ export class LocalTaskService implements TaskService {
     return of(null);
   }
 
-  private readTasks(): Task[] {
+  private readTasks(query?:string): Task[] {
+
     const tasks = localStorage.getItem(LocalTaskService.STORAGE_KEY);
-    return tasks ? JSON.parse(tasks) : [];
+    const tasksList = JSON.parse(tasks);
+    const searchResult = [];
+
+    if (tasks && !query) {
+      return tasksList;
+    } else {
+      tasksList.forEach((task)=> {
+        if (task.name.toLowerCase().includes(query.toLowerCase())) {
+          searchResult.push(task);
+        }
+      });
+      return searchResult;
+    }
   }
 
   private writeTasks(tasks: Task[]): void {
--- a/src/main/webapp/app/tasks/local-task.service.ts
+++ b/src/main/webapp/app/tasks/local-task.service.ts
@@ -40,13 +40,15 @@ export class LocalTaskService implements TaskService {
 
     if (tasks && !query) {
       return tasksList;
-    } else {
+    } else if (query) {
       tasksList.forEach((task)=> {
         if (task.name.toLowerCase().includes(query.toLowerCase())) {
           searchResult.push(task);
         }
       });
       return searchResult;
+    } else {
+      return [];
     }
   }
 
--- a/src/main/webapp/app/tasks/task-search/task-search.component.html
+++ b/src/main/webapp/app/tasks/task-search/task-search.component.html
@@ -1,0 +1,6 @@
+<form [formGroup]="taskSearch">
+  <mat-form-field class="search-bar">
+    <input (keyup)="searchQuery($event)" matInput placeholder="Search tasks">
+    <mat-icon matSuffix>search</mat-icon>
+  </mat-form-field>
+</form>
--- a/src/main/webapp/app/tasks/task-search/task-search.component.spec.ts
+++ b/src/main/webapp/app/tasks/task-search/task-search.component.spec.ts
@@ -1,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { TaskSearchComponent } from './task-search.component';
+
+describe('TaskSearchComponent', () => {
+  let component: TaskSearchComponent;
+  let fixture: ComponentFixture<TaskSearchComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ TaskSearchComponent ]
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(TaskSearchComponent);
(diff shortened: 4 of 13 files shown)


src/main/webapp/app/tasks/task-search/task-search.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(TaskSearchComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
```
djuerges: this test case is failing:

```
Failed: Template parse errors:
Can't bind to 'formGroup' since it isn't a known property of 'form'. ("<form [ERROR ->][formGroup]="taskSearch">
  <mat-form-field class="search-bar">
    <input (keyup)="searchQuery($even"): ng:///DynamicTestModule/TaskSearchComponent.html@0:6
'mat-icon' is not a known element:
1. If 'mat-icon' is an Angular component, then verify that it is part of this module.
2. If 'mat-icon' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message. ("ass="search-bar">
    <input (keyup)="searchQuery($event)" matInput placeholder="Search tasks">
    [ERROR ->]<mat-icon matSuffix>search</mat-icon>
  </mat-form-field>
</form>
"): ng:///DynamicTestModule/TaskSearchComponent.html@3:4
'mat-form-field' is not a known element:
1. If 'mat-form-field' is an Angular component, then verify that it is part of this module.
2. If 'mat-form-field' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message. ("<form [formGroup]="taskSearch">
  [ERROR ->]<mat-form-field class="search-bar">
    <input (keyup)="searchQuery($event)" matInput placeholder="Se"): ng:///DynamicTestModule/TaskSearchComponent.html@1:2
```

src/main/webapp/app/app.component.ts
```diff
@@ -26,6 +26,10 @@ export class AppComponent implements OnInit {
     this.tasks$ = this.taskService.getAll();
   }
 
+  search(query): void {
+    this.tasks$ = this.taskService.getAll(query);
```
djuerges: would prefer to add a new function when searching or e.g. something like` this.tasks$ = query ? this.taskService.searchTasks(query) : this.taskService.getAll();`

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -10,8 +10,8 @@ export class LocalTaskService implements TaskService {
 
   private static readonly STORAGE_KEY: string = 'tiny.tasks';
 
-  getAll(): Observable<Task[]> {
-    return of(this.readTasks());
+  getAll(query?:string): Observable<Task[]> {
+    return of(this.readTasks(query));
```
djuerges: like stated above, the main reason to add a new function for search functionality is, that it is a bit shady to rely on an optional query parameter in my opinion and doesn't make for a very good (as in readable and maintainable) code base, but more on that topic down below

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,9 +32,24 @@ export class LocalTaskService implements TaskService {
     return of(null);
   }
 
-  private readTasks(): Task[] {
+  private readTasks(query?:string): Task[] {
```
djuerges: like previously said, having the optional query does complicate things a bit in my opinion

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,9 +32,24 @@ export class LocalTaskService implements TaskService {
...
+      return tasksList;
+    } else if (query) {
+      tasksList.forEach((task)=> {
+        if (task.name.toLowerCase().includes(query.toLowerCase())) {
+          searchResult.push(task);
+        }
+      });
+      return searchResult;
+    } else {
+      return [];
```
djuerges: you don't need the else branch if you simply call `return searchResult;` at the end of the function, that way you can also save yourself the trouble to do the return in line 49 as well

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,9 +32,24 @@ export class LocalTaskService implements TaskService {
...
-  private readTasks(): Task[] {
+  private readTasks(query?:string): Task[] {
+
     const tasks = localStorage.getItem(LocalTaskService.STORAGE_KEY);
-    return tasks ? JSON.parse(tasks) : [];
+    const tasksList = JSON.parse(tasks);
+    const searchResult = [];
+
+    if (tasks && !query) {
+      return tasksList;
```
djuerges: by having a dedicated method for searching, this if branch wouldn't be necessary as you determine right away by the method call in the` app.component.ts` that you would have a valid query to search for

src/main/webapp/app/tasks/task-search/task-search.component.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { TaskSearchComponent } from './task-search.component';
+
+describe('TaskSearchComponent', () => {
```
djuerges: I would expect actual added test cases and not only the bare spec file that is generated by ng generate for instance

src/main/webapp/app/tasks/task-search/task-search.component.ts
```diff
@@ -0,0 +1,26 @@
+import {Component, EventEmitter, Output} from '@angular/core';
+import {FormControl, FormGroup, Validators} from '@angular/forms';
+
+/**
+ * Search field for tiny tasks.
+ */
+@Component({
+  selector: 'tiny-task-search',
+  templateUrl: './task-search.component.html',
+  styleUrls: ['./task-search.component.scss']
```
djuerges: it's always important to have `changeDetection: ChangeDetectionStrategy.OnPush` wherever possible for performance reasons

src/main/webapp/app/tasks/task-search/task-search.component.html
```diff
@@ -0,0 +1,6 @@
+<form [formGroup]="taskSearch">
```
djuerges: Why is it necessary to have a formGroup?

src/main/webapp/app/tasks/task-search/task-search.component.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+export class TaskSearchComponent {
+
+  taskSearch: FormGroup = new FormGroup({
+    queryInput: new FormControl('', Validators.required)
+  });
```
djuerges: I don't see the need to have a validator here, it looks more like a unnecessary copy&paste form the task-form.component.ts - adding also to this conclusion is the fact, that the validation does not even work are the is no `formControlName` with the name `queryInput`

src/main/webapp/app/tasks/task-search/task-search.component.html
```diff
@@ -0,0 +1,6 @@
...
+  <mat-form-field class="search-bar">
+    <input (keyup)="searchQuery($event)" matInput placeholder="Search tasks">
```
djuerges: need formControlName="queryInput" here when trying to add validation - it is missing and therefore broken now

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,9 +32,24 @@ export class LocalTaskService implements TaskService {
...
+    } else if (query) {
+      tasksList.forEach((task)=> {
```
djuerges: you can do this in a one-liner by filtering on tasksList and get rid of the` const searchResult` altogether

Issue #32, opened 2019-04-11
A user can attach files to a task


fynnfeldpausch:
As a user, I want to **attach files to a task** so that I can store additional information.

**Acceptance criteria**
- the user can upload a file
- the file is attached to the task
- the file can be downloaded

**Optional acceptance criteria**
- the user can see previews for images

**Additional information**
None.

PR #74, opened 2020-01-03
Task attachments added, test adjusted. 

aliilyas:
Solution for issue #32

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,7 @@ dependencies {
   implementation 'org.flywaydb:flyway-core'
   implementation 'net.rakugakibox.spring.boot:orika-spring-boot-starter:1.8.0'
   implementation 'org.postgresql:postgresql'
+  implementation 'commons-io:commons-io:2.6'
   compileOnly 'org.projectlombok:lombok'
   annotationProcessor 'org.projectlombok:lombok'
   testImplementation 'org.springframework.boot:spring-boot-starter-test'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,4 +9,6 @@ services:
       - POSTGRES_USER=tiny_task
       - POSTGRES_PASSWORD=demo123
     volumes:
-      - ./postgres/data:/var/lib/postgresql/data
+      - pgdata:/var/lib/postgresql/data
+volumes:
+  pgdata:
--- a/src/main/java/com/coyoapp/tinytask/dto/TaskResponse.java
+++ b/src/main/java/com/coyoapp/tinytask/dto/TaskResponse.java
@@ -14,4 +14,6 @@ public class TaskResponse {
   private String id;
 
   private String name;
+
+  private String fileName;
 }
--- a/src/main/java/com/coyoapp/tinytask/repository/DefaultFileRepository.java
+++ b/src/main/java/com/coyoapp/tinytask/repository/DefaultFileRepository.java
@@ -1,0 +1,50 @@
+package com.coyoapp.tinytask.repository;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+@Component
+public class DefaultFileRepository implements FileRepository {
+
+  @Value("${file.base.dir}")
+  private String filesBaseDir;
+
+  public Optional<Path> initFilesDir() {
+    Path path = Paths.get(filesBaseDir);
+    if (!Files.exists(path)) {
+      try {
+        path = Files.createDirectories(path);
+      } catch (IOException e) {
+        return Optional.empty();
+      }
+    }
+    return Optional.ofNullable(path);
+  }
+
+  public Optional<Path> saveFile(String name, byte[] content) {
+    try {
+      Path destination = Files.createFile(Paths.get(filesBaseDir, name));
+      try (FileOutputStream fos = new FileOutputStream(destination.toFile())) {
+        fos.write(content);
+      }
+      return Optional.ofNullable(destination);
+    } catch (Exception e) {
+      return Optional.empty();
+    }
+  }
+
+  public Optional<byte[]> getImageByName(String name) {
+    try {
+      return Optional.of(Files.readAllBytes(Paths.get(filesBaseDir, name)));
+    } catch (Exception e) {
+      return Optional.empty();
+    }
+  }
+}
--- a/src/main/java/com/coyoapp/tinytask/repository/FileRepository.java
+++ b/src/main/java/com/coyoapp/tinytask/repository/FileRepository.java
@@ -1,0 +1,12 @@
+package com.coyoapp.tinytask.repository;
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+public interface FileRepository {
+  Optional<Path> initFilesDir();
+
+  Optional<Path> saveFile(String name, byte[] content);
+
+  Optional<byte[]> getImageByName(String name);
+}
--- a/src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
+++ b/src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
@@ -1,0 +1,60 @@
+package com.coyoapp.tinytask.service;
+
+import com.coyoapp.tinytask.exception.FileException;
+import com.coyoapp.tinytask.repository.FileRepository;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.FilenameUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.web.multipart.MultipartFile;
+
+import javax.annotation.PostConstruct;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.Optional;
+
+@Slf4j
+@Component
+@RequiredArgsConstructor
+public class DefaultFileService implements FileService {
+
+  @Autowired
+  private FileRepository fileRepository;
+
+  @PostConstruct
+  private void initImageRootFolder() {
+    Path path = fileRepository.initFilesDir()
+      .orElseThrow(() -> new RuntimeException("File base directory not accesssible."));
+    log.debug("initImageRootFolder(imageBaseDir={})", path.toAbsolutePath());
+  }
+
+  public String saveFile(Optional<MultipartFile> file) {
+    String fileName = null;
+    if (file.isPresent() && !file.get().isEmpty()) {
+      try {
+        MultipartFile multipartFile = file.get();
+        fileName = wrapFileName(multipartFile);
+        fileRepository.saveFile(fileName, multipartFile.getBytes()).orElseThrow(() -> new FileException("File saving failed."));
+      } catch (IOException e) {
+        throw new FileException("File reading failed.");
+      }
+    }
+    return fileName;
+  }
+
+  public byte[] getFile(String fileName) {
+    return fileRepository.getImageByName(fileName).orElseThrow(() -> new FileException("File reading error."));
+  }
+
+  private String wrapFileName(MultipartFile multipartFile) {
+    String origName = multipartFile.getOriginalFilename();
+    return new StringBuilder()
+      .append(FilenameUtils.getBaseName(origName).replaceAll("\\W", "-")) // replace all non word chars with -
+      .append(Instant.now().getNano())
+      .append('.')
+      .append(FilenameUtils.getExtension(origName))
+      .toString();
+  }
+}
--- a/src/main/java/com/coyoapp/tinytask/service/DefaultTaskService.java
+++ b/src/main/java/com/coyoapp/tinytask/service/DefaultTaskService.java
@@ -5,12 +5,15 @@ import com.coyoapp.tinytask.dto.TaskRequest;
 import com.coyoapp.tinytask.dto.TaskResponse;
 import com.coyoapp.tinytask.exception.TaskNotFoundException;
 import com.coyoapp.tinytask.repository.TaskRepository;
-import java.util.List;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import ma.glasnost.orika.MapperFacade;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
+import org.springframework.web.multipart.MultipartFile;
+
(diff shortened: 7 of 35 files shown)


docker-compose.yml
```diff
@@ -9,4 +9,6 @@ services:
       - POSTGRES_USER=tiny_task
       - POSTGRES_PASSWORD=demo123
     volumes:
-      - ./postgres/data:/var/lib/postgresql/data
+      - pgdata:/var/lib/postgresql/data
```
GeBeater: not part of task #32. what is the reason for this change and why volumes are used over mounts?

build.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
   implementation 'org.flywaydb:flyway-core'
   implementation 'net.rakugakibox.spring.boot:orika-spring-boot-starter:1.8.0'
   implementation 'org.postgresql:postgresql'
```
GeBeater: commons-io:commons-io or rather org.apache.commons.io.FilenameUtils would no longer be required by applying the commentary listed below

src/main/java/com/coyoapp/tinytask/dto/TaskResponse.java
```diff
@@ -14,4 +14,6 @@
   private String id;
 
   private String name;
+
```
GeBeater: due to the adjustments in the DefaultFileService it is more a unique file identifier instead of the origin filename

src/main/resources/application.yml
```diff
@@ -3,12 +3,26 @@ spring:
...
       ddl-auto: validate
+    properties:
+      hibernate:
+        temp:
+          use_jdbc_metadata_defaults: false
 
 logging:
   level:
     com.coyoapp.tinytask: DEBUG
+
```
GeBeater: backend does not start out of the box due to /storage/.. is read-only. fixed it by create directory /tmp/tiny-task on local machine. an adjustment of the README that the path must be readable or changed would be great

src/main/java/com/coyoapp/tinytask/repository/DefaultFileRepository.java
```diff
@@ -0,0 +1,50 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+@Component
+public class DefaultFileRepository implements FileRepository {
+
+  @Value("${file.base.dir}")
+  private String filesBaseDir;
+
```
GeBeater: I would simplify and rename the method like below.

```
  public Path createFileBaseDirectoryIfNotExists() {
    try {
      return Files.createDirectories(Paths.get(filesBaseDir));
    } catch (IOException e) {
      throw new IllegalArgumentException("Could not create file base dir", e);
    }
  }
```

this optional makes the code unnecessarily complicated. especially an exception is thrown anyway if the directory could not be created (see DefaultFileService.initImageRootFolder())

src/main/java/com/coyoapp/tinytask/repository/DefaultFileRepository.java
```diff
@@ -0,0 +1,50 @@
...
+    if (!Files.exists(path)) {
+      try {
+        path = Files.createDirectories(path);
+      } catch (IOException e) {
+        return Optional.empty();
+      }
+    }
+    return Optional.ofNullable(path);
+  }
+
```
GeBeater: same argumentation regarding optional here. I would simplify the code like below.

```
  public Path saveFile(String name, byte[] content) {
    try {
      Path destination = Files.createFile(Paths.get(filesBaseDir, name));
      try (FileOutputStream fos = new FileOutputStream(destination.toFile())) {
        fos.write(content);
        return destination;
      }
    } catch (Exception e) {
      throw new FileException("Could not create file");
    }
  }
```

src/main/java/com/coyoapp/tinytask/repository/DefaultFileRepository.java
```diff
@@ -0,0 +1,50 @@
...
+      Path destination = Files.createFile(Paths.get(filesBaseDir, name));
+      try (FileOutputStream fos = new FileOutputStream(destination.toFile())) {
+        fos.write(content);
+      }
+      return Optional.ofNullable(destination);
+    } catch (Exception e) {
+      return Optional.empty();
+    }
+  }
+
```
GeBeater: same argumentation for optionals here. would simplify the code (no example here; inspire by examples above).

src/main/java/com/coyoapp/tinytask/repository/FileRepository.java
```diff
@@ -0,0 +1,12 @@
+package com.coyoapp.tinytask.repository;
+
+import java.nio.file.Path;
+import java.util.Optional;
+
```
GeBeater: the repositories in this package are spring data repository. could be confuse for someone.

src/main/java/com/coyoapp/tinytask/repository/FileRepository.java
```diff
@@ -0,0 +1,12 @@
...
+public interface FileRepository {
```
GeBeater: optionals can removed, see comments above

src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Optional;
+
+@Slf4j
+@Component
+@RequiredArgsConstructor
+public class DefaultFileService implements FileService {
+
+  @Autowired
+  private FileRepository fileRepository;
+
```
GeBeater: due to comment above, method can be simplified by ...

```
  @PostConstruct
  private void initImageRootFolder() {
    fileRepository.createFileBaseDirectoryIfNotExists();
  }
```

src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
```diff
@@ -0,0 +1,60 @@
...
+  @PostConstruct
+  private void initImageRootFolder() {
+    Path path = fileRepository.initFilesDir()
+      .orElseThrow(() -> new RuntimeException("File base directory not accesssible."));
+    log.debug("initImageRootFolder(imageBaseDir={})", path.toAbsolutePath());
+  }
+
```
GeBeater: due to comment above, method can be simplified by ...

```
  public String saveFile(MultipartFile file) {
    String fileName = UUID.randomUUID().toString();
    try {
      fileRepository.saveFile(fileName, file.getBytes());
    } catch (Exception e) {
      throw new FileException("File saving failed.");
    }
    return fileName;
  }
```

src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
```diff
@@ -0,0 +1,60 @@
...
+        MultipartFile multipartFile = file.get();
+        fileName = wrapFileName(multipartFile);
+        fileRepository.saveFile(fileName, multipartFile.getBytes()).orElseThrow(() -> new FileException("File saving failed."));
+      } catch (IOException e) {
+        throw new FileException("File reading failed.");
+      }
+    }
+    return fileName;
+  }
+
```
GeBeater: due to comment above, method can be simplified by ...

```
  public byte[] getFile(String fileName) {
    return fileRepository.getImageByName(fileName);
  }
```

src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
```diff
@@ -0,0 +1,60 @@
...
+  public byte[] getFile(String fileName) {
+    return fileRepository.getImageByName(fileName).orElseThrow(() -> new FileException("File reading error."));
+  }
+
```
GeBeater: this looks like, that the filename was used to generate a unique file identifier. thus I would simplify the code by using UUID.randomUUID().toString() instead

src/main/java/com/coyoapp/tinytask/service/DefaultTaskService.java
```diff
@@ -20,13 +23,16 @@
 public class DefaultTaskService implements TaskService {
 
   private final TaskRepository taskRepository;
+  private final FileService fileService;
   private final MapperFacade mapperFacade;
 
   @Override
   @Transactional
```
GeBeater: optionals as parameter is generally not the best idea. i ask you to do some research on this yourself, so that I can skip the argumentation here.

src/main/java/com/coyoapp/tinytask/service/TaskService.java
```diff
@@ -2,11 +2,14 @@
 
 import com.coyoapp.tinytask.dto.TaskRequest;
 import com.coyoapp.tinytask.dto.TaskResponse;
+import org.springframework.web.multipart.MultipartFile;
+
 import java.util.List;
+import java.util.Optional;
 
 public interface TaskService {
```
GeBeater: see "optionals as parameter" comment above

src/main/java/com/coyoapp/tinytask/web/TaskController.java
```diff
@@ -27,9 +24,9 @@
   private final TaskService taskService;
 
   @PostMapping
```
GeBeater: javax.validation.constraints for @Valid no longer works due to usage of @RequestParam

src/main/java/com/coyoapp/tinytask/repository/DefaultFileRepository.java
```diff
@@ -0,0 +1,50 @@
+package com.coyoapp.tinytask.repository;
```
GeBeater: no tests for DefaultFileRepository available. (does not reflect skill sheet rating (especially for TTD) 😉)

src/test/java/com/coyoapp/tinytask/service/DefaultFileServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+
+import java.util.Optional;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.when;
+
```
GeBeater: only getFile of DefaultFileService is tested. test coverage does not reflect skill sheet rating (especially for TTD) 😉

src/test/java/com/coyoapp/tinytask/service/DefaultFileServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+  @InjectMocks
+  private DefaultFileService objectUnderTest;
+
+  @Rule
+  public TemporaryFolder baseFileDir = new TemporaryFolder();
+
+  @Test
+  public void shouldReadFromRepo() {
+    //when
```
GeBeater: filename not considered for test

src/test/java/com/coyoapp/tinytask/service/DefaultFileServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+
+import java.util.Optional;
+
```
GeBeater: hamcrest assertThat and mockito matchers instead of AssertJ (see already existing tests)

src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
```diff
@@ -0,0 +1,60 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.Optional;
+
+@Slf4j
+@Component
+@RequiredArgsConstructor
+public class DefaultFileService implements FileService {
+
```
GeBeater: constructor injection should be preferred (see other components) instead of field injection

src/test/java/com/coyoapp/tinytask/web/TaskControllerTest.java
```diff
@@ -50,6 +45,29 @@ public void shouldCreateTask() throws Exception {
       .andExpect(jsonPath("$.name", is(name)));
   }
 
+  @Test
```
GeBeater: I would prefer to extend the existing shouldCreateTask test instead of adding the shouldCreateTaskWithFile (for this testing purpose)

src/main/webapp/app/app.module.ts
```diff
@@ -4,9 +4,10 @@ import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
```
GeBeater: I skipped the frontend review remarks in general. unfortunately there are also some things to note

src/main/resources/application.yml
aliilyas: Did not pay attention to that, tried to make less config dependent solution.

src/main/java/com/coyoapp/tinytask/web/TaskController.java
aliilyas: Roger that! missed that.

src/main/java/com/coyoapp/tinytask/service/DefaultTaskService.java
aliilyas: It depends on project coding conventions. There is no blocker argument not to use that.

src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
aliilyas: Nope, UUIDs instead of readable file names are harder to read for users. Just pretend you downloaded attachments from a couple of tasks and trying to identify them in the downloads folder. (Sorry that's my nightmare XD )

src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
aliilyas: "File reading failed." is wrong should be "File reading input MultipartFile failed.", `multipartFile.getBytes()` can throw IOException. 
`fileRepository.saveFile` can throw a bulk of different exceptions, which are simplified to single optional to make raise FileException fluent one-liner.

src/main/java/com/coyoapp/tinytask/service/DefaultFileService.java
aliilyas: Roger that!

src/main/java/com/coyoapp/tinytask/repository/FileRepository.java
aliilyas: Creating another repository package will be more confusing. By default repository is a data access layer, in this case, implementation of a repository can be changed to use database storage, sending file to some cloud storage.

Issue #27, opened 2019-04-05
A user can set a due date


fynnfeldpausch:
As a user, I want to **set a due date** for my tasks so that I can start working on my urgent tasks first.

**Acceptance criteria**
- the user can set a due date for a task
- the task list is sorted by due date
  - tasks with a due date are sorted in a descending order
  - tasks without a due date are listed below

**Optional acceptance criteria**
- none

**Additional information**
None

PR #78, opened 2020-01-25
A user can set a due date #27

mohamedelkashif:
I implemented the due date as described in the opened issue #27 .
I modified in scss files to be more responsiveness.


src/main/webapp/app/tasks/task-form/task-form.component.ts
```diff
@@ -14,17 +14,25 @@ import { TaskService } from '../task.service';
...
+    name: new FormControl('', Validators.required),
+    dueDate: new FormControl(''),
+    time: new FormControl('')
   });
 
   constructor(@Inject('TaskService') private taskService: TaskService) { }
 
   onSubmit(): void {
-    this.taskService.create(this.taskForm.value.name).subscribe(task => {
+    if(this.taskForm.value.dueDate && this.taskForm.value.time){
```
patrickbeck1337: I would use the patchValue method here and move the setting of the time to date into a specific private method.  Like:
`  onSubmit(): void {
    if (this.taskForm.value.dueDate && this.taskForm.value.time) {
      this.taskForm.patchValue({dueDate: this.setTimeToDate(this.taskForm.value.dueDate, this.taskForm.value.time)});
    }
    this.taskService.create(this.taskForm.value.name, this.taskForm.value.dueDate).subscribe(task => {
      this.created.emit(task);
      this.taskForm.reset();
    });
  }

  private setTimeToDate(date: Date, time: string): string {
    const hour = +time.split(':')[0];
    const minute = +time.split(':')[1];
    date.setHours(hour, minute);
    return date.toISOString();
  }`

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -20,7 +20,7 @@ export interface TaskService {
    * @param name the task's name
    * @returns an `Observable` holding the created task
    */
-  create(name: string): Observable<Task>;
+  create(name: string, dueDate: Date): Observable<Task>;
```
patrickbeck1337: dueDate should be optional in my opinion.

PR #80, opened 2020-02-29
A user can mark a task as done #29

mohsammarraie:
Implemented the mentioned above feature. (only frontend part).


cryve:
Thanks for your effort! Lets discuss your solution in the Code Review Talk :)

PR #97, opened 2021-02-27
filter task feature by Stas

SSlippa:


src/main/webapp/app/tasks/filter-pipe/filter-task.pipe.ts
```diff
@@ -0,0 +1,26 @@
...
+  name: 'filterTask',
+  pure: false
+})
+export class FilterTaskPipe implements PipeTransform {
+
+  transform(tasks: Task[], text: string): Task[] {
+    if (text) {
+      if (text.includes('AND')) {
+        const splittedText = text.split('AND');
+        return tasks.filter(task => splittedText.every(part => task.name.toLowerCase().includes(part.trim().toLowerCase())));
```
djuerges, line 14: should extract `task.name.toLowerCase().includes(part.trim().toLowerCase()` to a single method because it is re-used three times

src/main/webapp/app/tasks/filter-pipe/filter-task.pipe.ts
```diff
@@ -0,0 +1,26 @@
...
+      } else if (text.includes('OR')) {
+        const splittedText = text.split('OR');
+        return tasks.filter(task => splittedText.some(part => task.name.toLowerCase().includes(part.trim().toLowerCase())));
+      } else {
+        return tasks.filter(task => task.name.toLowerCase().includes(text.toLowerCase()));
```
djuerges, line 19: not using `trim()` here means the search might behave slightly different when no search operator is provided as whitespaces are also considered, not sure if this is intentional

src/main/webapp/app/tasks/filter-pipe/filter-task.pipe.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Task } from '../task';
+
+@Pipe({
+  name: 'filterTask',
+  pure: false
+})
+export class FilterTaskPipe implements PipeTransform {
+
+  transform(tasks: Task[], text: string): Task[] {
+    if (text) {
```
djuerges, line 11: will trigger also on whitespaces, should check for empty strings

src/main/webapp/app/tasks/task-search/task-search.component.ts
```diff
@@ -0,0 +1,45 @@
+import { Component, EventEmitter, OnDestroy, OnInit, Output } from '@angular/core';
+import { FormControl, FormGroup, Validators } from '@angular/forms';
+import { LocalTaskService } from 'app/tasks/local-task.service';
+import { Subscription } from 'rxjs';
+import { distinctUntilChanged } from 'rxjs/operators';
+
+@Component({
+  selector: 'tiny-task-search',
+  templateUrl: './task-search.component.html',
+  styleUrls: ['./task-search.component.scss']
```
djuerges, line 10: would be nice to use the change detection strategy `On.Push`

src/main/webapp/app/app.component.html
```diff
@@ -4,12 +4,13 @@
       <div class="logo">
         <img src="assets/images/logos/coyo/logo-coyo-inversed-nav-hd.png" alt="COYO">
       </div>
+      <tiny-task-search (searchEvent)="searchTask($event)"></tiny-task-search>
```
djuerges, line 7: UI wise I probably would have placed this below the header and above the listed items

src/main/webapp/app/app.component.ts
```diff
@@ -27,4 +29,8 @@ export class AppComponent implements OnInit {
   deleted(): void {
     this.tasks$ = this.taskService.getAll();
   }
+
+  searchTask(text: string): void {
+    this.searchedText = text;
```
djuerges, line 34: What made you choose this over a direct service call?

src/main/webapp/app/tasks/filter-pipe/filter-task.pipe.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    // given
+    const filterText = 'Hana AND Sarah';
+
+    // then
+    const filtered = filterPipe.transform(arrToBeFiltered, filterText);
+
+    // expected
+    expect(filtered).toEqual([{id: '4', name: 'Hana Joe Sarah'}]);
+  });
+});
```
djuerges, line 64: 👍🏼  for tests

src/main/webapp/app/tasks/task-search/task-search.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    text: new FormControl('')
+  });
+
+  subscriptions = new Subscription();
+  @Output() searchEvent = new EventEmitter<string>();
+
+  constructor() { }
+
+  ngOnInit(): void {
+    this.bindEvents();
```
djuerges, line 23: no real need to extract this into a seperate method, also _bindEvents_ doesn't tell a whole lot about what is happening here

src/main/webapp/app/tasks/task-search/task-search.component.ts
SSlippa: But here I don't have any Inputs so I don't see any reason for On.Push

src/main/webapp/app/app.component.ts
SSlippa: If I subscribe on some Subject with using service in this component, I think I should have used manually change detection to update the list, here my Input is changed every time and components were pretty close one to another

src/main/webapp/app/tasks/task-search/task-search.component.ts
SSlippa: I used to use this way usually in components with big forms, so maybe it should be called bindSaveEvents and generally, there could be a lot of logic like subscribe on each control in the form and each group and its controls and some logic inside these subs...
usually for live save on different controls

src/main/webapp/app/tasks/filter-pipe/filter-task.pipe.ts
SSlippa: Right, fixed

src/main/webapp/app/tasks/filter-pipe/filter-task.pipe.ts
SSlippa: Right, fixed

src/main/webapp/app/tasks/filter-pipe/filter-task.pipe.ts
SSlippa: Right, fixed

PR #101, opened 2021-03-03
Front end - Feature Update - A user can mark a task as done

sreekanthkylm:
screencapture-localhost-4200-2021-03-05-10_29_05

**Front end - Feature Update - A user can mark a task as done**

- the user can mark a task as done
- the task list is sorted by done status
- tasks that are not done go to the top of the list
- tasks that are done go to the bottom of the list
- the user can clear the list of done tasks
- the user can set the task status with more details ("In Progress", "Blocked")

### DIFF
--- a/src/main/webapp/app/tasks/default-task.service.ts
+++ b/src/main/webapp/app/tasks/default-task.service.ts
@@ -24,8 +24,8 @@ export class DefaultTaskService implements TaskService {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
   }
 
-  updateStatus(id: string, status: string): Observable<void> {
-    return this.http.get<void>(this.baseUrl + '/tasks/' + id);
+  updateStatus(id: string, status: string): Observable<Task[]> {
+    return this.http.put<Task[]>(this.baseUrl + '/tasks/' + `/${id}` + `?status=${status}`, null);
   }
 
   clearCompletedTasks(): Observable<void> {
--- a/src/main/webapp/app/tasks/default-task.service.ts
+++ b/src/main/webapp/app/tasks/default-task.service.ts
@@ -13,7 +13,7 @@ export class DefaultTaskService implements TaskService {
   }
 
   create(name: string): Observable<Task> {
-    return this.http.post<Task>(this.baseUrl + '/tasks', {name: name} as Task);
+    return this.http.post<Task>(this.baseUrl + '/tasks', { name: name } as Task);
   }
 
   delete(id: string): Observable<void> {
@@ -22,5 +22,13 @@ export class DefaultTaskService implements TaskService {
 
   getAll(): Observable<Task[]> {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
+  }
+
+  updateStatus(id: string, status: string): Observable<void> {
+    return this.http.get<void>(this.baseUrl + '/tasks/' + id);
+  }
+
+  clearCompletedTasks(): Observable<void> {
+    return this.http.get<void>(this.baseUrl + '/tasks/');
   }
 }
--- a/src/main/webapp/app/tasks/local-task.service.ts
+++ b/src/main/webapp/app/tasks/local-task.service.ts
@@ -14,9 +14,9 @@ export class LocalTaskService implements TaskService {
     return of(this.readTasks());
   }
 
-  create(name: string): Observable<Task> {
+  create(name: string, status: string): Observable<Task> {
     const tasks = this.readTasks();
-    const task = {id: uuid(), name};
+    const task = { id: uuid(), name, status };
     tasks.push(task);
     this.writeTasks(tasks);
     return of(task);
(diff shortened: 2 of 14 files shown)


src/main/webapp/app/tasks/default-task.service.ts
```diff
@@ -23,4 +23,12 @@ export class DefaultTaskService implements TaskService {
   getAll(): Observable<Task[]> {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
   }
+
+  updateStatus(id: string, status: string): Observable<void> {
+    return this.http.get<void>(this.baseUrl + '/tasks/' + id);
```
djuerges, line 28: you wouldn't update the status with a get

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,6 +32,25 @@ export class LocalTaskService implements TaskService {
...
+      }
+      return task;
+    });
+    this.writeTasks(updated_tasks);
+    return of(null);
+  }
+
+  clearCompletedTasks(): Observable<void> {
+    const tasks = this.readTasks();
+    const clearedTasks = tasks.filter(task => task.status !== 'done');
```
djuerges, line 49: to make it a bit cleaner, one could have done using or defining some sort of types/typing instead of strings

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,6 +32,25 @@ export class LocalTaskService implements TaskService {
...
+  updateStatus(id: string, status: string): Observable<void> {
+    const tasks = this.readTasks();
+    const updated_tasks = tasks.map(task => {
+      if (task.id === id) {
+        task.status = status;
+      }
+      return task;
+    });
+    this.writeTasks(updated_tasks);
+    return of(null);
```
djuerges, line 44: why return observale of null?

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,47 @@
 <mat-list data-cy="task-list">
-  <mat-list-item *ngFor="let task of tasks" class="mat-elevation-z1">
-    <mat-icon mat-list-icon>assignment</mat-icon>
-    <h4 mat-line>{{task.name}}</h4>
-    <button mat-icon-button aria-label="Delete task" color="primary">
-      <mat-icon aria-label="Delete task" (click)="delete(task)">delete</mat-icon>
-    </button>
-  </mat-list-item>
-</mat-list>
+  <section *ngIf="pendingTasks?.length>0" class="completed-lists">
```
djuerges, line 2: what do you think of just using *ngIf="pendingTasks" and not checking the length?

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,47 @@
...
+        <mat-icon matListIcon>done</mat-icon>
+      </button>
+      <button mat-icon-button aria-label="Update task status" color="primary"
+        (click)="updateStatus(task,'in progress')">
+        <mat-icon matListIcon *ngIf="task.status==='in progress'" [ngStyle]="{'color':'blue'}">pending</mat-icon>
+        <mat-icon matListIcon *ngIf="task.status!=='in progress'">pending</mat-icon>
+      </button>
+      <button mat-icon-button aria-label="Update task status" color="primary" (click)="updateStatus(task,'blocked')">
+        <mat-icon matListIcon *ngIf="task.status==='blocked'" [ngStyle]="{'color':'red'}">block</mat-icon>
+        <mat-icon matListIcon *ngIf="task.status!=='blocked'">block</mat-icon>
```
djuerges, line 23: seems like code dulication, what could you imagine to change to make it more generic and re-usable?

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,47 @@
...
+      <button mat-icon-button aria-label="Delete task" color="primary" (click)="delete(task)">
+        <mat-icon aria-label="Delete task">delete</mat-icon>
+      </button>
+      <button mat-icon-button aria-label="Update task status" color="primary" (click)="updateStatus(task,'done')">
+        <mat-icon matListIcon>done</mat-icon>
+      </button>
+      <button mat-icon-button aria-label="Update task status" color="primary"
+        (click)="updateStatus(task,'in progress')">
+        <mat-icon matListIcon *ngIf="task.status==='in progress'" [ngStyle]="{'color':'blue'}">pending</mat-icon>
+        <mat-icon matListIcon *ngIf="task.status!=='in progress'">pending</mat-icon>
```
djuerges, line 19: I would probably prefer a dropdown over having so much different icons with regards to changing the same property

src/main/webapp/app/tasks/local-task.service.ts
djuerges, line 44: why return observable of null?

src/main/webapp/app/tasks/default-task.service.ts
sreekanthkylm: :) Missed that. Updated with put.

src/main/webapp/app/tasks/task-list/task-list.component.html
sreekanthkylm: We return an empty array if there is no value in local storage. That's why there is an extra check of length

src/main/webapp/app/tasks/task-list/task-list.component.html
sreekanthkylm: Updated with conditions in ngstyle.

src/main/webapp/app/tasks/task-list/task-list.component.html
sreekanthkylm: Updated with a dropdown feature

src/main/webapp/app/tasks/local-task.service.ts
sreekanthkylm: Updated. if api need to return the updated task . since we are using localstorage returned the full list.

src/main/webapp/app/tasks/local-task.service.ts
sreekanthkylm: Updated with property check

PR #102, opened 2021-03-03
Implementing the search features

uinoushi:
Task Done:-

1. Adding a shared module
2. Adding search component and search pipe
3. Implementing the search component inside app component
4. Covering the unit and integration tests.
5. Using cypress to create regression tests.

Todo:-

1. Optional AC

### DIFF
--- a/src/main/cypress/integration/app.spec.ts
+++ b/src/main/cypress/integration/app.spec.ts
@@ -4,4 +4,28 @@ context('App', () => {
   beforeEach(() => {
     cy.visit('/');
   });
+
+  describe('Search task', () => {
+    it('should create 3 tasks and filter them by a needle term', () => {
+      cy.get('[data-cy=task-input').type('Task 1');
+      cy.get('[data-cy=send').click();
+  
+      cy.get('[data-cy=task-input').type('Task 2');
+      cy.get('[data-cy=send').click();
+  
+      cy.get('[data-cy=task-input').type('Task 3');
+      cy.get('[data-cy=send').click();
+  
+      cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 3);
+  
+      cy.get('[data-cy=search-input').type('Task 3');
+      cy.get('[aria-label=Clear]').should('be.visible');
+      cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 1);
+  
+      cy.get('[aria-label=Clear]').click();
+  
+      cy.get('[data-cy=search-input').type('Test');
+      cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 0);
+    })
+  })
 });
--- a/src/main/cypress/integration/app.spec.ts
+++ b/src/main/cypress/integration/app.spec.ts
@@ -23,9 +23,13 @@ context('App', () => {
       cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 1);
   
       cy.get('[aria-label=Clear]').click();
+      cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 3);
   
       cy.get('[data-cy=search-input').type('Test');
       cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 0);
+
+      cy.get('[aria-label=Clear]').click();
+      cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 3);
     })
   })
 });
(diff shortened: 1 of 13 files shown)


src/main/cypress/integration/app.spec.ts
```diff
@@ -4,4 +4,28 @@ context('App', () => {
...
+      cy.get('[aria-label=Clear]').should('be.visible');
+      cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 1);
+  
+      cy.get('[aria-label=Clear]').click();
+  
+      cy.get('[data-cy=search-input').type('Test');
+      cy.get('[data-cy=task-list').find('.mat-list-item').should('have.length', 0);
+    })
+  })
 });
```
djuerges, line 31: nice one!

src/main/webapp/app/shared/search/search.component.html
```diff
@@ -0,0 +1,26 @@
...
+      class="search-input"
+    />
+    <mat-icon matSuffix *ngIf="!value">search</mat-icon>
+    <button
+      mat-button
+      *ngIf="value"
+      matSuffix
+      mat-icon-button
+      aria-label="Clear"
+      (click)="value = ''"
```
djuerges, line 21: this won't lead to tasks being reloaded, it will just clear the value

src/main/webapp/app/shared/search/search.component.scss
```diff
@@ -0,0 +1,29 @@
...
+  .search-field {
+    margin-left: auto;
+    width: 200px;
+  }
+
+  .mat-icon {
+    color: grey;
+  }
+
+  ::ng-deep {
```
djuerges, line 14: this is oudated, what would you do instead?

src/main/webapp/app/app.component.html
```diff
@@ -9,7 +9,9 @@
   </mat-toolbar>
   <section class="content">
     <div class="wrapper">
-      <tiny-task-list [tasks]="tasks$ | async" (deleted)="deleted($event)"></tiny-task-list>
+      <tiny-search [(value)]="term"></tiny-search>
```
djuerges, line 12: what made you chose this binding over using an eventEmitter here like e.g. (search)?

src/main/webapp/app/shared/search/search.component.html
uinoushi: Good catch :)
I fix it.

src/main/webapp/app/shared/search/search.component.scss
uinoushi: I replaced the **ng-deep** by turning the view encapsulation off but it is also not a best practice either. 
IMHO it would be better that we move the logic to global SCSS file to create some SCSS mixins. WDYT?

src/main/webapp/app/app.component.html
uinoushi: At the moment, my changes are specifically for the search feature.
Later, in case having more time, my plan was to transform the search to a global component form-input-text.
The objective behind it is to encapsulate the mat-form-field in one single component to be used by the other ones. 
Please check the sample below:

```
// shared/form-input-text/form-input-text.ts

import { Component, forwardRef, Input } from '@angular/core';
import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';

/**
 * Displays regular text input.
 */
@Component({
    selector: 'tiny-form-input-text',
    templateUrl: './form-input-text.component.html',
    styleUrls: ['./form-input-text.component.scss'],
    providers: [
        {
            provide: NG_VALUE_ACCESSOR,
            useExisting: forwardRef(() => FormInputTextComponent),
            multi: true,
        },
    ],
})
export class FormInputTextComponent implements ControlValueAccessor {
    @Input() placeholder = '';
    @Input() label = '';
    @Input() icon = '';
    @Input() error = '';
    isTouched = false;
    value = '';

    registerOnChange(onChange: (value: string) => void): void {
        this.onChange = onChange;
    }

    registerOnTouched(onTouched: () => void): void {
        this.onTouched = onTouched;
    }

    writeValue(value: string): void {
        this.value = value || '';
    }

    pushChanges(value: string) {
        this.writeValue(value);
        this.onChange(this.value);
    }

    pushTouches() {
        this.isTouched = true;
        this.onTouched();
    }

    protected onTouched = () => {};
    protected onChange = (value: string) => {};
}
```

PR #104, opened 2021-03-05
search feature with test cases

jinuccherian:


jinuccherian:
user can search for tasks [#28](https://github.com/mindsmash/tiny-tasks/issues/28)

PR #107, opened 2021-03-12
A user can set a due date

f1m4uk:
Enhancements:
  * empty state illustration is added to be shown, when user has no tasks yet (better UX);
  * "delete" action was removed (and replaced with 3 vertical dots) from the list due to two reasons:
    * destructive action is a main action of a list item without any confirmation (bad UX);
    * as more the application will grow as more UX will suffer: list item cannot contain more than 1 secondary action;
  * the whole list item is now clickable for better UX and it triggers context menu with additional actions to appear;
  * destructive action is highlighted with "error" color for better UX;
  * task dialog form is introduced for adding (e.g. for future FAB button)/editing tasks:
    * it is fully responsive: full-screen on mobile devices, free size for the rest;
    * user won't be able to accidentally close it if the child form is not pristine, bottom sheet with cancellation approval will be opened;
    * progress bar is added to inform a user about loading state, which is great for poor network devices;
  * native datetime picker is used, because Angular Material one has no time picking feature yet;
  * WCAG compliant (only the new functionality).
 
Main task:
  * user can set "Due date" for any task inside the dialog form.
  * the tasks list is sorted from the backend side. However, that's not a good idea. There must be an Angular Material Sort Header component above the list or "Sort by" dropdown.
  * user can set the time as well.

Suggestions (there are a lot of things to fix even in such a small app, I just have no more time and don't know if it's worth):
  * main field to add tasks is not responsive. Switch to mobile landscape or even iPhone 5S portrait to be sure - layout breaks;
  * (click) events should not be listened to on mat-icons...
  * aria-label doesn't have to be duplicated twice on button and mat-icon. You can just add even single matTooltip;
  * `color` attribute is not working;
  * it is not so easy to manage the loading state of a list, because of Input-Output communication between the list and parent component. Tasks could be retrieved by the list itself.
  * use camelCase Directives always (where it's possible), dash-case is deprecated a long time ago;
  * migrate to Angular 11, it will give HMR feature, which is awesome. (Also strict TS types);
  * the main logo alt contains capital letters, so screen readers pronounce it wrongly. Alt should describe the image, btw. "Coyo brand logo" would be better;
  * ...

Tech notes:
  * automated tests are added both backend and frontend:
    * please, note that bottom sheets and dialogs are very tricky to test (they're not part of Angular rendering flow).
  * according to Material Specs all buttons should be capitalized, however some screen readers might read that letter by letter, which is wrong (But I tested via ChromeVox, works fine).

P.S. Some styles are hardcoded, because it's super time consuming to start building customly themed app (and even Angular Material doesn't have everything out of the box for that).
P.S.S. I know, that's a "dummy" project :)


klui:
Thanks for your extensive pull request, and the detailed explanation above. I'll give you a summary of my findings an impression below.

**General impression:**
Nice to have tackled fullstack task.
Extensive solution with lots of extra not required functionality - editing in dialog, preventing unwanted deletion through bottom sheet.
Detailed (UX) enhancement ideas, looking into future features and extensions - and already implemented (e.g. good approach introducing context menu when another action was introduced).

**Details to be challenged, but not so relevant for this task:**
Making the whole element clickable although there is an explicit context menu (which is not clickable on its own) is questionable UX.
As accessibility is especially mentioned as a goal in the PR comments, one might wonder if the dialog, with another bottom sheet opening, is really a good accessible solution.

**Main task - adding the possibility to add a due date to a task with sorting:**
Overall good woking solution - backend and frontend.
Some important functionality details are missing though:
- Form validation on the date input field would have been a real plus.
- Ordering by due date not working on frontend-only version.
- Generally a focus on this main task (maybe also with the frontend sorting as suggested by yourself) would have been appreciated

**Code style/presentation**
Empty files or functions can be deleted, for example scss files that are not used or empty ngOnInit functions.
Docs for added functions would have been nice.
There are many linting errors. If the project readme refers to a linting yarn task, it would be expected to run without errors.
Some newly introduced methods are missing tests (e.g. TaskListComponent.openTaskDialogForm).
Some tests are not actually testing the method functionality, but only the mocking (although I know it’s hard testing angular material dialogs).
The TaskListComponent became very crowded with code defining how the TaskFormDialog should be rendered. This should have been extracted for separation of concerns and reusability of the TaskFormDialog (e.g. for also adding a new task). Additionally using the breakpointObserver introduces extra complexity and possible errors (unsubscribing hot observables) where a simple CSS media query solution would have been cleaner and sufficient for almost all styles implemented here.

src/main/resources/db/migration/V1__Initial_Setup.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE task (
     id VARCHAR(36) CONSTRAINT task_id_pkey PRIMARY KEY,
     name VARCHAR (128) NOT NULL,
+    due_date TIMESTAMP WITH TIME ZONE,
```
klui, line 4: Please never change existing migration scripts, but always add a new one.

src/main/webapp/app/cancellation-approval-bottom-sheet/cancellation-approval-bottom-sheet.component.html
```diff
@@ -0,0 +1,10 @@
+<mat-action-list>
+  <button mat-list-item (click)="bottomSheetRef.dismiss(true)">
```
klui, line 2: also this tiny bit of logic would be nicer in the component ts file

src/main/webapp/app/app.module.ts
```diff
@@ -1,34 +1,44 @@
...
     HttpClientModule
   ],
   providers: [
-    {provide: BASE_URL, useValue: 'http://localhost:8080'},
-    {provide: 'TaskService', useClass: (environment.useLocalStorage) ? LocalTaskService : DefaultTaskService}
+    { provide: BASE_URL, useValue: 'http://localhost:8080' },
+    { provide: 'TaskService', useClass: (environment.useLocalStorage) ? LocalTaskService : DefaultTaskService }
+  ],
+  entryComponents: [
+    CancellationApprovalBottomSheetComponent
```
klui, line 41: with Ivy there is no need for this in the entryComponents anymore

src/main/webapp/app/cancellation-approval-bottom-sheet/cancellation-approval-bottom-sheet.component.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+      .compileComponents();
+  });
+
+  beforeEach(() => {
+    matBottomSheet = TestBed.inject(MatBottomSheet);
+    bottomSheetSpy = spyOn(matBottomSheet, 'open').and.returnValue(bottomSheetRefSpyObj);
+  });
+
+  it('should open', async () => {
+    matBottomSheet.open(MatBottomSheetHarness);
```
klui, line 26: Actually you are never testing the CancellationApprovalBottomSheetComponent but only if your mocks work.

src/main/webapp/app/tasks/local-task.service.spec.ts
```diff
@@ -52,4 +52,13 @@ describe('LocalTaskService', () => {
     expect(localStorage.getItem).toHaveBeenCalled();
     expect(localStorage.setItem).toHaveBeenCalled();
   });
+  
+  it('should update task in local storage', () => {
+    // when
+    taskService.update({ id: id, name: 'Buy a kilo of Ritter Sport chocolate' });
+
+    // then
+    expect(localStorage.getItem).toHaveBeenCalled();
```
klui, line 61: as the expects are the same as above this can't really test the functionality of this single method

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -12,17 +19,89 @@ import { TaskService } from '../task.service';
...
+  }
 
   delete(task: Task): void {
     this.taskService.delete(task.id).subscribe(() => {
       this.deleted.emit(task);
     });
   }
+
+  openTaskFormDialog(task: Task) {
+    const subscription = combineLatest([this.isHandset$, this.isWeb$]).subscribe(info => {
```
klui, line 64: this should also be unsubscribed on destroy of the component

klui (commented):
Just some detailed code lines to point to examples from what I already wrote in my general review.

f1m4uk:
Thank you for awesome advices. Let me answer some of your thoughts.

>Making the whole element clickable although there is an explicit context menu (which is not clickable on its own) is questionable UX.

Agree with you. Material specs don't mention anything about that. Lists item can contain both primary and secondary actions. However, that must be discussed in team - whether the list is going to have two actions or just single one. As long as I've noticed that "primary" action is not interactable I decided to proceed with Angular Material action lists component (entire row is clickable). IMHO, very awesome UX, especially, for mobile users. Again, for this particular application only, because I would expect the item to be expanded on clicking/tapping (or redirected to /:id) for other apps (e.g. Gmail).
>As accessibility is especially mentioned as a goal in the PR comments, one might wonder if the dialog, with another bottom sheet opening, is really a good accessible solution.

Well, Angular Material did a really great job for that. I've checked focus and screen reader behavior. Everything is pretty interactable and obvious (assuming I'm blind).
>Empty files or functions can be deleted, for example scss files that are not used or empty ngOnInit functions.

Hmm, never heard about that. I always use cli generate, so don't pay much attention to empty styles. Nevertheless, thanks for that suggestion.
>Docs for added functions would have been nice.

I know that docs are really important thing. However, I'm trying my best to name methods and variables as much obvious as it's possible. Especially, working with .ts simplifies life a lot. But I'm not an enemy of docs :)
>There are many linting errors. If the project readme refers to a linting yarn task, it would be expected to run without errors.

Yeah, good catch. I missed that.
>The TaskListComponent became very crowded with code defining how the TaskFormDialog should be rendered. This should have been extracted for separation of concerns and reusability of the TaskFormDialog (e.g. for also adding a new task). Additionally using the breakpointObserver introduces extra complexity and possible errors (unsubscribing hot observables) where a simple CSS media query solution would have been cleaner and sufficient for almost all styles implemented here.

Yes, the breakpointObserver adds a lot of code. However, it has huge advantage over CSS media queries - HTML won't be rendered at all, unlike CSS only hides things.
>Form validation on the date input field would have been a real plus

Is not default input type validation enough for such a trivial sample? There was no requirement to make `dueDate` not earlier than today or anything similar.
>Ordering by due date not working on frontend-only version.

As I've mentioned in my initial comment, current ordering is not scalable at all. There must be unified mechanism to sort tasks (material sort bar or "sort by" button). Also, frontend sorting is not good in case if you have pagination or lazy-loading. So, frontend sorting is just a bear's service.
>Generally a focus on this main task (maybe also with the frontend sorting as suggested by yourself) would have been appreciated

That's a bit offensive :D
The REQs are done. Anything else was just my intention to show small piece of my skills/creativeness etc.
I mean, the main task can be done within 1.5-2 hours. It wouldn't show most of the skills.

In summary: should I fix all of those problems for you to be able to merge? Or just wait?

src/main/webapp/app/cancellation-approval-bottom-sheet/cancellation-approval-bottom-sheet.component.html
f1m4uk: I did that intentionally. Could you tell me why is that really tiny thing must be put into .ts? Really curious.

klui:
Thanks for your answers already. There is no need for you to fix any of those things mentioned - they were mainly a feedback for you. We will close the pull request without merging (to keep the original project for other applicants to work with).
Sorry if some of my notes came across offensive - it sure wasn't meant that way. I just saw you are very ambitious and I wanted to give as detailed feedback as possible (I know some stuff is quite picky and sure things open for discussion or personal taste).
Please wait for Miriam to contact you about the further process. I think we will have the chance to discuss some parts of your solution later on.

PR #109, opened 2021-03-19
A user can set a due date, both ways backend and locally

moni12345:


src/main/webapp/app/tasks/task-form/task-form.component.ts
```diff
@@ -1,32 +1,41 @@
...
 })
 export class TaskFormComponent {
 
   @Output() created: EventEmitter<Task> = new EventEmitter();
+  today = new Date();
 
   taskForm: FormGroup = new FormGroup({
-    name: new FormControl('', Validators.required)
+    name: new FormControl('', Validators.required),
+    duedate: new FormControl('', Validators.required)
```
klui, line 25: From the requirements ("tasks with a due date go to the top of the list, tasks without a due date go to the bottom of the list") the due date should have stayed optional.

src/main/webapp/app/tasks/task-form/task-form.component.ts
```diff
@@ -1,32 +1,41 @@
 import { ChangeDetectionStrategy, Component, EventEmitter, Inject, Output } from '@angular/core';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 
+// import moment from 'moment';
```
klui, line 4: those kind of comments - especially in the wrong place - are more confusing than of any help

src/main/webapp/app/tasks/task-form/task-form.component.spec.ts
```diff
@@ -33,39 +33,39 @@ describe('TaskFormComponent', () => {
...
-    component.taskForm.setValue({name: 'My task'});
+    component.taskForm.setValue({name: 'My task', duedate: '18 Mar 2021'});
     expect(component.taskForm.invalid).toBe(false);
   });
 
   it('should create a task', () => {
     // given
-    component.taskForm.setValue({name: 'My task'});
-    taskService.create.and.returnValue(of({id: 'id', name: 'My task'}));
+    component.taskForm.setValue({name: 'My task', duedate: '18 Mar 2021'});
```
klui, line 42: Your main logic is within the tested file. You should at least adapt the test value of component.taskForm.value to something that is realistically returned from the browser, so that your formatting logic with moment.js is tested. Even better would be an additional test which tests only the correct date transformation with moment in the onSubmit method

src/main/webapp/app/tasks/task-form/task-form.component.html
```diff
@@ -1,6 +1,12 @@
 <form [formGroup]="taskForm" (ngSubmit)="onSubmit()">
   <mat-form-field>
     <input type="text" formControlName="name" matInput placeholder="Add a new tiny task…" data-cy="task-input">
-    <mat-icon matSuffix aria-label="Add task"(click)="onSubmit()">send</mat-icon>
   </mat-form-field>
+  <mat-form-field appearance="fill">
```
klui, line 5: this appearances gives it some nasty layout and background shade, would have gone without it

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -34,7 +34,20 @@ export class LocalTaskService implements TaskService {
 
   private readTasks(): Task[] {
     const tasks = localStorage.getItem(LocalTaskService.STORAGE_KEY);
-    return tasks ? JSON.parse(tasks) : [];
+    const taskJson = JSON.parse(tasks); 
+    taskJson.sort(comp);
```
klui, line 38: I'd prefer an inline function for this. But more important: This is some newly added functionality which should be tested.

src/main/resources/db/migration/V1__Initial_Setup.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE task (
     id VARCHAR(36) CONSTRAINT task_id_pkey PRIMARY KEY,
     name VARCHAR (128) NOT NULL,
+    duedate DATE,
```
klui, line 4: Just a hint: Please don't edit existing migration scripts when working with flyway, but just add a new one.

src/main/java/com/coyoapp/tinytask/web/TaskController.java
```diff
@@ -25,17 +29,18 @@
...
     log.debug("createTask(createTask={})", taskRequest);
     return taskService.createTask(taskRequest);
   }
 
   @GetMapping
-  public List<TaskResponse> getTasks() {
+  public List<Task> getTasks() {
     log.debug("getTasks()");
-    return taskService.getTasks();
+    return taskRepository.getTasks();
```
klui, line 43: Also just a hint as this is backend code: Do not access repositories directly from the controller, but use at least a device layer in between.

src/main/java/com/coyoapp/tinytask/web/TaskController.java
```diff
@@ -25,17 +29,18 @@
 class TaskController {
 
   private final TaskService taskService;
-
+  @Autowired
+  private TaskRepository taskRepository;
```
klui, line 33: Also just a hint as this is backend code: Adding this breaks the backend tests, you would have to adapt them.

src/main/java/com/coyoapp/tinytask/repository/TaskRepository.java
```diff
@@ -2,6 +2,14 @@
 
 import com.coyoapp.tinytask.domain.Task;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+
+import java.util.List;
 
 public interface TaskRepository extends JpaRepository<Task, String> {
+
+  @Query(value = "SELECT * FROM task t ORDER BY duedate ASC",
```
klui, line 11: Also just a hint as this is backend code: A custom query is not needed here, you could have called the existing JpaRepository method findAll(Sort...) with a sort parameter.

src/main/java/com/coyoapp/tinytask/dto/TaskRequest.java
```diff
@@ -1,10 +1,17 @@
 package com.coyoapp.tinytask.dto;
 
+import javax.persistence.Temporal;
```
klui, line 3: Please remove unused imports

src/main/java/com/coyoapp/tinytask/dto/TaskRequest.java
```diff
@@ -14,5 +21,8 @@
 
   @NotEmpty
   private String name;
+  @JsonFormat(pattern = "dd MMM yyyy")
```
klui, line 24: Using this pattern, I get a backend error: "JSON parse error: Cannot deserialize value of type `java.util.Date` from String "25 Mar 2021": expected format "dd MMM yyyy"
Apparently moment does really have another formatting (especially when it comes to capitals), but it would have been safer to only go with numbers like dd-MM-yyyy anyway.

klui (commented):
General impression: Short solution with a focus on the task. Unfortunately some basic code style requirements are not met (have a look at the many errors from yarn lint), none of the new functionality is tested.

src/main/java/com/coyoapp/tinytask/repository/TaskRepository.java
moni12345: So basically I tried JpaRepository method to sort in a way that all duadate with null values comes after all non null values, which is not working in a way I'm expecting, so that's why I used the custom query.

src/main/java/com/coyoapp/tinytask/dto/TaskRequest.java
moni12345: I don't know but it is working fine on my system.

PR #110, opened 2021-03-21
PR for Task Done issue 

LievStudio:
I worked on the delete task issue.
As a side note, I wanted to add a Material Select to the project's imports to handle one  of the optional acceptance challenges but I could never compile the code after adding such import. Adding a native select element looked too... ugly for me so I decided not to include it for now.
Regards,

Jorge


src/main/webapp/app/app.component.html
```diff
@@ -2,14 +2,49 @@
...
+          alt="COYO"
+        />
       </div>
-      <tiny-task-form (created)="created($event)"></tiny-task-form>
+      <tiny-task-form (created)="created()"></tiny-task-form>
     </div>
   </mat-toolbar>
-  <section class="content">
+  <section class="content flex-container">
+    <span style="display: flex"
```
djuerges, line 14: please _at all cost_ don't style inline

src/main/webapp/app/app.component.html
```diff
@@ -2,14 +2,49 @@
...
+          <button
+            *ngIf="doneTasks.length > 0"
+            mat-raised-button
+            color="warn"
+            (click)="clearDone()"
+          >
+            Clear Done Tasks
+          </button></span
+        >
+        <hr />
```
djuerges, line 41: not a fan of the horizontal line here, in my opinion it can lead to the perception that the started tasks are somehow connected to the done tasks

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -1,37 +1,55 @@
...
       this.writeTasks(tasks);
     }
     return of(null);
   }
 
+  toggleStatus(id: string): Observable<void> {
+    const tasks = this.readTasks();
+    tasks.forEach((task) =>
+      task.id === id
+        ? task.status === "Started"
```
djuerges, line 38: would be cleaner to export an interface and type the status

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,20 @@
 <mat-list data-cy="task-list">
   <mat-list-item *ngFor="let task of tasks" class="mat-elevation-z1">
-    <mat-icon mat-list-icon>assignment</mat-icon>
-    <h4 mat-line>{{task.name}}</h4>
+    <mat-icon mat-list-icon (click)="done(task)">assignment</mat-icon>
+    <h4 mat-line>{{ task.name }}</h4>
+    <h5 mat-line style="font-weight: 500">
```
djuerges, line 5: again, inline styling

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,20 @@
...
+      Status:
+      <span
+        [ngStyle]="{ color: task.status === 'Done' ? 'Green' : 'darkblue' }"
```
djuerges, line 8: I would set a class depending on the status and then apply the coloring there

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -1,28 +1,43 @@
...
  * A list of tiny tasks.
  */
 @Component({
-  selector: 'tiny-task-list',
-  templateUrl: './task-list.component.html',
-  styleUrls: ['./task-list.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush
+  selector: "tiny-task-list",
+  templateUrl: "./task-list.component.html",
+  styleUrls: ["./task-list.component.scss"],
```
djuerges, line 20: this breaks our linting: `ERROR: 8:8    quotemark  " should be '`

src/main/webapp/app/tasks/task.ts
```diff
@@ -4,4 +4,5 @@
 export interface Task {
   id: string;
   name: string;
+  status: string;
```
djuerges, line 7: if not type is used you could limit the inputs to "Started" | "Done" and so on

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -1,37 +1,55 @@
...
+          ? (task.status = "Done")
+          : (task.status = "Started")
+        : null
```
djuerges, line 41: this is pretty hard to understand, in my opinion for each and then match the id and do the assignment is not a very good idea, you would rather find the index and set it directly or delegate to a method or there are lot of different other options to choose from, but two nested ternary operators?

src/main/webapp/app/app.component.html
```diff
@@ -2,14 +2,49 @@
...
-      <tiny-task-form (created)="created($event)"></tiny-task-form>
+      <tiny-task-form (created)="created()"></tiny-task-form>
     </div>
   </mat-toolbar>
-  <section class="content">
+  <section class="content flex-container">
+    <span style="display: flex"
+      ><h2>Click</h2>
+      <mat-icon class="title-icon">assignment</mat-icon>
+      <h2>to toggle status of Task</h2></span
```
djuerges, line 17: a poor UI decision in my opinion because a good UI would be intuitively understandable to the user without an explanation how to use it

src/main/webapp/app/app.component.ts
```diff
@@ -1,30 +1,64 @@
...
-import { Task } from './tasks/task';
-import { TaskService } from './tasks/task.service';
+import { Task } from "./tasks/task";
 
 @Component({
-  selector: 'tiny-root',
-  templateUrl: './app.component.html',
-  styleUrls: ['./app.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush
+  selector: "tiny-root",
```
djuerges, line 14: breaks our linting

PR #111, opened 2021-03-23
Feature/task done

atillabaspinar2:
"done" (boolean) field is added to Task class. 

Frontend new features:
 - User can change done to true by clicking on a checkbox. 
 - Checkbox is removed afterward. 
 - A "done" task is moved down in the list. 
 - Implemented local storage implementation for the "done" field. 
 - Task list returns sorted (by "done" field) after each request. 

Backend:
- Update controller is implemented in the backend.
- Migration created for "done" field in "task" model.


src/main/webapp/app/app.component.ts
```diff
@@ -1,20 +1,24 @@
...
-import { TaskService } from './tasks/task.service';
+import { Task } from "./tasks/task";
+import { TaskService } from "./tasks/task.service";
 
 @Component({
-  selector: 'tiny-root',
-  templateUrl: './app.component.html',
-  styleUrls: ['./app.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush
+  selector: "tiny-root",
```
djuerges, line 13: unfortunately this breaks our linting

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -1,30 +1,41 @@
...
-    const task = {id: uuid(), name};
+    const task = { id: uuid(), name };
     tasks.push(task);
     this.writeTasks(tasks);
     return of(task);
   }
 
+  update(task: Task): Observable<Task> {
+    const tasks = this.readTasks();
+    const index = tasks.findIndex((t) => t.id == task.id);
```
djuerges, line 27: probably should use === just to be on the safe side

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -1,28 +1,84 @@
...
+        this.deleted.emit(task);
+        this.snackBar.open("Task deleted!");
+      },
+      (error) => {
+        this.snackBar.open("Task deletion failed!");
+      }
+    );
+  }
+
+  toggleStatus(task: Task) {
```
djuerges, line 51: missing return type definition

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -1,28 +1,84 @@
...
+        this.updated.emit();
+        this.snackBar.open("Task status updated!");
+      },
+      (error) => {
+        this.snackBar.open("Task status update failed!");
+      }
+    );
+  }
+
+  deleteAllDone() {
```
djuerges, line 64: missing return type definition

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -1,28 +1,84 @@
...
+        this.deleted.emit();
+        this.snackBar.open("All done tasks deleted!");
+      },
+      (error) => {
+        this.snackBar.open("Batch delete failed!");
+      }
+    );
+  }
+
+  drop(event: CdkDragDrop<Task[]>) {
```
djuerges, line 81: missing return type definition

src/main/webapp/app/tasks/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { Task } from "./task";
+
+export function sortTasksByStatus(tasks: Task[]) {
```
djuerges, line 3: missing return type definition

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,24 @@
...
+
+
+<mat-list cdkDropList (cdkDropListDropped)="drop($event)" data-cy="task-list">
+  <mat-list-item *ngFor="let task of tasks" class="mat-elevation-z1 task-list-item" [ngClass]="{'task-done': task.done}"
+    cdkDrag>
+    <mat-icon mat-list-icon *ngIf="task.done">assignment_turned_in</mat-icon>
+    <mat-icon mat-list-icon *ngIf="!task.done">assignment</mat-icon>
     <h4 mat-line>{{task.name}}</h4>
+    <button mat-icon-button aria-label="Toggle status" color="primary">
+      <mat-icon aria-label="Toggle task status" (click)="toggleStatus(task)" *ngIf="!task.done">check_box_outline_blank
```
djuerges, line 16: as you only show the icon on !task.done, there will be still be the button present and a hover effect when you mouse over it, would rather do this on the button

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,24 @@
...
-    <mat-icon mat-list-icon>assignment</mat-icon>
+<div class="task-list-menu mat-elevation-z1">
+
+  <button mat-icon-button aria-label="Delete all done tasks" color="primary" matTooltip="Delete all done tasks">
+    <mat-icon color="warn" aria-label="Delete all done tasks" (click)="deleteAllDone(task)">remove_done</mat-icon>
+  </button>
+</div>
+
+
+<mat-list cdkDropList (cdkDropListDropped)="drop($event)" data-cy="task-list">
```
djuerges, line 9: nice that you implemented the drag & drop but what's the functional use? a user cannot use it to mark something as done

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -1,28 +1,84 @@
...
+      }
+    );
+  }
+
+  toggleStatus(task: Task) {
+    task.done = !task.done;
+    this.taskService.update(task).subscribe(
+      () => {
+        this.updated.emit();
+        this.snackBar.open("Task status updated!");
```
djuerges, line 56: bonus points next time for also adding the new status ;)

src/main/webapp/app/app.component.ts
atillabaspinar2: These syntax changes are because of my prettier config. Maybe it is good idea to save the prettier config in the app.

src/main/webapp/app/tasks/local-task.service.ts
atillabaspinar2: Definitely. This is a typo that I made accidentally..

src/main/webapp/app/tasks/task-list/task-list.component.html
atillabaspinar2: I was aware of this. But I see no point in drag&drop later. Because I already had a checkbox. I was a bit confused of requirements. It mentions two tables but then also mentions of moving up and down a task. In any case, I should have removed drag & drop code (maybe in the next commit :)

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -1,28 +1,84 @@
...
  */
 @Component({
-  selector: 'tiny-task-list',
-  templateUrl: './task-list.component.html',
-  styleUrls: ['./task-list.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush
+  selector: "tiny-task-list",
+  templateUrl: "./task-list.component.html",
+  styleUrls: ["./task-list.component.scss"],
+  encapsulation: ViewEncapsulation.None,
```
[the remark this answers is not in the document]
atillabaspinar2: It is because dragged item loses its style. Because the dragged item is attached to the body, instead of the component. 
Maybe a better solution can be found when further investigated.

src/main/webapp/app/tasks/task-list/task-list.component.ts
atillabaspinar2: Yes, right. For consistency, functions should have a return type. Even though these functions are used only in the template and nowhere still, this is not an excuse.

src/main/webapp/app/tasks/task-list/task-list.component.ts
atillabaspinar2: Right, a lot of changes are possible, but please beware that I only had 3 hours in total :)

src/main/webapp/app/tasks/task-list/task-list.component.ts
atillabaspinar2: Again, right. I will add return types of functions next time ;)

src/main/webapp/app/tasks/task-list/task-list.component.html
atillabaspinar2: You mean "delete all done" button? Ah yes, this button may be removed once clicked until there is a new "done" task. But no big harm and there is no extra request sent to the backend.

src/main/webapp/app/tasks/utils.ts
atillabaspinar2: I think I will definitely remember to add return types to all functions from now on :)

src/main/webapp/app/app.component.ts
atillabaspinar2: You mean the double quotes? Right. I do not know how this happened. I should have reviewed the commit before pushing.

src/main/webapp/app/tasks/local-task.service.ts
atillabaspinar2: Definitely! I missed that.

src/main/webapp/app/tasks/task-list/task-list.component.ts
atillabaspinar2: Definitely. I missed that.

src/main/webapp/app/tasks/task-list/task-list.component.ts
atillabaspinar2: You are right.  It does not break the code but yes, we are writing TS not JS.

src/main/webapp/app/tasks/task-list/task-list.component.html
atillabaspinar2: There is room to improve the behavior for sure. Here there is no harm to have the "delete all done" button but for sure, this can be improved as you said.

src/main/webapp/app/tasks/task-list/task-list.component.html
atillabaspinar2: I lated decided that I do not need a drag&drop behavior. The code is there for an example :) (Joking, I should have removed that code)

src/main/webapp/app/tasks/task-list/task-list.component.ts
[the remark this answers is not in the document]
atillabaspinar2: ViewEncapsulation.None? 
Because the dragged element is attached to the body and loses its style that is specified in the component.

src/main/webapp/app/tasks/task-list/task-list.component.ts
atillabaspinar2: Alright! :)

atillabaspinar2 (commented):
Thanks for the evaluation!

src/main/webapp/app/tasks/task-list/task-list.component.html
atillabaspinar2: I was just reviewing the code again. I just noticed that *ngIf is on the icon. It should be on the button of course.

atillabaspinar2: Btw, I am not used to GitHub but rather Gitlab. That's why I did not understand where your comments point to, at the first sight.

PR #116, opened 2021-04-25
Task melissa

melissaceleste:


.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "arrowParens": "avoid",
+  "semi": false,
```
djuerges, line 3: this breaks our linting when you run `yarn lint` 😬

src/main/webapp/app/tasks/task.ts
```diff
@@ -2,6 +2,9 @@
...
  */
 export interface Task {
-  id: string;
-  name: string;
+  id: string
+  name: string
+  status: string
 }
+
+//here i would prefer using an boolean with: isChecked: true/false - but this attempt worked easier for the beginning
```
djuerges, line 10: Totally agree, boolean is way better and I think it might also have made things easier a bit because the sorting get's an issue and assigning the correct status value requires more logic.

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,48 @@
...
+    <button mat-icon-button aria-label="Done task">
+      <mat-icon
+        aria-label="mark task as done"
+        (change)="setIsDone($event)"
+        [ngStyle]="{ color: task.status === 'done' ? 'green' : 'primary' }"
+        >done</mat-icon
+      >
     </button>
   </mat-list-item>
+  <button mat-button aria-label="Delete task" (click)="deleteDoneTasks(task)">
```
djuerges, line 28: this doesn't need the task to be passed and indeed the task is not part of the signature/a paramater

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -1,28 +1,62 @@
...
+    })
+  }
+
+  setIsDone(task: Task): void {
+    this.taskService
+      .setIsDone(task.id)
+      .subscribe(() => this.statusChanged.emit(task))
+  }
+
+  deleteDoneTasks(task: Task): void {
```
djuerges, line 44: the task does not need to be passed to remove all status = done entries from the array later on, so I would change the siganture to  `deleteDoneTasks(task: Task): void`

src/main/webapp/app/tasks/local-task.service.spec.ts
```diff
@@ -1,55 +1,72 @@
...
+    expect(localStorage.getItem).toHaveBeenCalled()
+    expect(localStorage.setItem).toHaveBeenCalled()
+  })
+
+  it('should safe the done status to local storage', () => {
+    // when
+    taskService.setIsDone('done')
+
+    // then
+    expect(localStorage.setItem).toHaveBeenCalled()
```
djuerges, line 61: maybe forgot a `getItem` here?

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -1,43 +1,62 @@
...
+    }
+    return of(null)
+  }
+
+  setIsDone(id: string): Observable<void> {
+    const tasks = this.readTasks()
+    tasks.forEach(task =>
+      task.id === id ? (task.status = 'done') : (task.status = 'to do')
+    )
+    // here or in the task-list.component.ts I would sort the tasks by their status
```
djuerges, line 39: right track, but how would you do it?

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -1,32 +1,57 @@
...
+   */
+  setIsDone(status: string): Observable<void>
+
+  /**
+   * Delete the done tasks.
+   *
+   * @param id the ID of which is done
+   * @returns an empty `Observable`
+   */
+  deleteDoneTasks(id: string): Observable<void>
```
djuerges, line 46: as mentioned before, this works without knowing a specific id

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,48 @@
...
+      <mat-icon aria-label="Delete task" (click)="delete(task)"
+        >delete</mat-icon
+      >
+    </button>
+    <label>
+      <input
+        aria-label="mark task as done"
+        type="checkbox"
+        [color]="'primary'"
+        [checked]="task.done"
```
djuerges, line 15: does not work because you made `done` string-typed, instead would need to write [checked]="task.done === 'done'"

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,48 @@
...
-      <mat-icon aria-label="Delete task" (click)="delete(task)">delete</mat-icon>
+      <mat-icon aria-label="Delete task" (click)="delete(task)"
+        >delete</mat-icon
+      >
+    </button>
+    <label>
+      <input
+        aria-label="mark task as done"
+        type="checkbox"
+        [color]="'primary'"
```
djuerges, line 14: using [] in angular means it is an @Input to the component, however this is not true here so `color="primary"` because then the input also does not need to be string-typed

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -1,32 +1,57 @@
...
-  delete(id: string): Observable<void>;
+  delete(id: string): Observable<void>
+
+  /**
+   * Mark the task as done.
+   *
+   * @param id the ID of which is done
+   * @returns an empty `Observable`
+   */
+  setIsDone(status: string): Observable<void>
```
djuerges, line 38: in `setIsDone` of local-task.service you use the parameter as id 🤔  must be rather `setIsDone(id: string): Observable<void>`

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,9 +1,48 @@
...
+      />
+    </label>
+    <button mat-icon-button aria-label="Done task">
+      <mat-icon
+        aria-label="mark task as done"
+        (change)="setIsDone($event)"
+        [ngStyle]="{ color: task.status === 'done' ? 'green' : 'primary' }"
+        >done</mat-icon
+      >
     </button>
```
djuerges, line 26: go with the button, way better ui 🤞🏼

.prettierrc
melissaceleste: oh. Good to know! :D

src/main/webapp/app/tasks/task.ts
melissaceleste: Yes. After reading your following comment, I definitly would use a boolean next time. My boolean-attempt just caused errors. But after your review I understood much more, so I think I'll understand these errors a way better now.

src/main/webapp/app/tasks/task-list/task-list.component.html
melissaceleste: So this should be filled with the "done.task" or maybe just be empty?

src/main/webapp/app/tasks/local-task.service.spec.ts
melissaceleste: Yes I forgot that. Thank you!

src/main/webapp/app/tasks/task.service.ts
melissaceleste: Okay I think I could have solved this easier, when I would have used the boolean-attempt. Thanks for your comment!

src/main/webapp/app/tasks/task-list/task-list.component.html
melissaceleste: Yes careless mistake 🙈 But this would work, if I would use the boolean-attempt or?

src/main/webapp/app/tasks/task-list/task-list.component.html
melissaceleste: Okay that's a good point. Because this was something, I haven't completely understood.

src/main/webapp/app/tasks/local-task.service.ts
melissaceleste: I thought about somethink like  
```
const a = tasks.status = 'done'
const b = tasks.status = 'to do'
tasks.sort(a, b){return a - b}
```
But this don't work. So I think it should be easier with a boolean. Because then it's easier to use the localeCompare(). Something like this  `.sort((x,y) => x.task.status === y.task.status ? 0 : x ? 1 : -1))`

What would you prefer?

melissaceleste:
Thank you for your review! I would like to modify my code and try the boolean-attempt again :) Should I send it to you or how is the process now? I hope I didn't failed with my "easy status-string-attempt". 😬

PR #117, opened 2021-04-25
Task victoria

VictoriaWika:


src/main/webapp/app/tasks/local-task.service.spec.ts
```diff
@@ -6,7 +6,8 @@ import { Task } from './task';
 describe('LocalTaskService', () => {
   const id = 'de4f576e-d1b5-488a-8c77-63d4c8726909';
   const name = 'Doing the do!';
-  const mockTask = `{"id":"${id}","name":"${name}"}`;
+  const completed = false;
+  const mockTask = `{"id":"${id}","name":"${name}", "completed":"${completed}"}`;
```
djuerges, line 10: the value of  mockTask completed is assigned as string and will be interpreted as such, making '**should return tasks from local storage**' fail with _Expected 'false' to equal false._

src/main/webapp/app/tasks/task-list/task-list.component.spec.ts
```diff
@@ -48,9 +48,31 @@ describe('TaskListComponent', () => {
...
+    component.delete({id: 'id', name: 'My task', completed: false});
 
     // then
-    expect(deleteEmitter).toHaveBeenCalledWith({id: 'id', name: 'My task'});
+    expect(deleteEmitter).toHaveBeenCalledWith({id: 'id', name: 'My task', completed: false});
+  });
+
+  it('should mark a task as completed', () => {
+    // given
+    taskService.togglecomplete.and.returnValue(of(null));
```
djuerges, line 59: this test is failing, as yu're new to the Angular world you probably have overlooked the fact that you need to mock the called method in line 13 like

`taskService = jasmine.createSpyObj('taskService', ['delete', 'togglecomplete']);`

src/main/webapp/app/tasks/task-list/task-list.component.spec.ts
```diff
@@ -48,9 +48,31 @@ describe('TaskListComponent', () => {
...
+    // when
+    component.togglecomplete({id: 'id', name: 'My task', completed: false});
+
+    // then
+    expect(taskService.togglecomplete).toHaveBeenCalledWith('id');
+  });
+
+  it('should delete all completed tasks', () => {
+    // given
+    taskService.clearcomplete.and.returnValue(of(null));
```
djuerges, line 70: similar to above, this test is failing, as you need to mock the called method in line 13 like

`taskService = jasmine.createSpyObj('taskService', ['delete', 'clearcomplete']);`

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -11,12 +11,12 @@ export class LocalTaskService implements TaskService {
   private static readonly STORAGE_KEY: string = 'tiny.tasks';
 
   getAll(): Observable<Task[]> {
-    return of(this.readTasks());
+    return of(this.readTasks().sort((x,y) => x.completed === y.completed ? 0 : x? 1 : -1));
```
djuerges, line 14: This does not work. 😕  As you've correctly seen one of the acceptance criteria is to sort the tasks by done, to the bottom. The express is almost complete but make it work but one crucial detail is missing. For the second ternary operator you are missing the property you want to sort against.

`x.completed `

making this work with

`return of(this.readTasks().sort((x,y) => x.completed === y.completed ? 0 : x.completed ? 1 : -1));`

I guess think went throught because there was no test case to make sure the sorting works as expected.

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,6 +32,23 @@ export class LocalTaskService implements TaskService {
     return of(null);
   }
 
+  togglecomplete(id:string): Observable<void> {
+    const tasks = this.readTasks();
+    tasks.map(task => {
+      if ( task.id === id) {
+      task.completed = !task.completed;
+      this.writeTasks(tasks);
```
djuerges, line 40: text intendation +2 characters for line 39 + 40

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,6 +32,23 @@ export class LocalTaskService implements TaskService {
...
+      }
+    })
+    return of(null);
+  }
+
+  clearcomplete(): Observable<void> {
+    const tasks = this.readTasks();
+    this.writeTasks(tasks.filter(task => task.completed === false));
+    return of(null);
```
djuerges, line 49: Do you think it would make sense to return the task here? `of(null)` was used in delete because there simply is no task to return anymore after it is gone.

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,6 +32,23 @@ export class LocalTaskService implements TaskService {
...
 
+  togglecomplete(id:string): Observable<void> {
+    const tasks = this.readTasks();
+    tasks.map(task => {
+      if ( task.id === id) {
+      task.completed = !task.completed;
+      this.writeTasks(tasks);
+      }
+    })
+    return of(null);
```
djuerges, line 43: see comment below

src/main/webapp/app/tasks/local-task.service.ts
VictoriaWika: You are right, totally missed it here!

src/main/webapp/app/tasks/task-list/task-list.component.spec.ts
VictoriaWika: I have added both functions, thanks!

src/main/webapp/app/tasks/local-task.service.spec.ts
VictoriaWika: That's right! Somehow I didn't notice... I've tried to fix the issue with the tests, but I can't seem to find the problem. Do you have a suggestion?

PR #118, opened 2021-05-03
29 user can mark task as done

shafeequeonline:
I have implemented the functions in the acceptance criteria. I have used the Shared module logic for the checkbox component. So we can reuse the same in our application. Used until-destroy package, it is a good way to unsubscribe from observables when the component destroyed.

- the user can mark a task as done 
- the task list is sorted by done status
- tasks that are not done go to the top of the list
- tasks that are done go to the bottom of the list
- the user can clear the list of done tasks


src/main/webapp/app/shared/components/checkbox/checkbox.component.html
```diff
@@ -0,0 +1 @@
+<mat-checkbox (click)="pushChanges('')">{{name}}</mat-checkbox>
```
TimSielemann, line 1: Why did you decided to create a checkbox component when its only wrapping the mat-checkbox? :)

Also the click handler seems odd as it does not push any changes but an empty string.

src/main/webapp/app/shared/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,18 @@
...
+})
+export class CheckboxComponent {
+  @Input() name: string;
+  @Input() click: string;
+
+  pushChanges(value: string): void {
+    this.onChange();
+  }
+
+  protected onChange = () => {};
```
TimSielemann, line 16: what is this lambda function for?

src/main/webapp/app/shared/shared.module.ts
```diff
@@ -0,0 +1,16 @@
...
+
+@NgModule({
+  declarations: [CheckboxComponent],
+  imports: [
+    CommonModule,
+    MatCheckboxModule
+  ],
+  exports: [CheckboxComponent]
+})
+export class SharedModule { }
```
TimSielemann, line 16: We do it the same here at coyo but of course with more than one shared module :)

src/main/webapp/app/tasks/default-task.service.spec.ts
```diff
@@ -61,4 +61,16 @@ describe('DefaultTaskService', () => {
     // finally
     req.flush({});
   });
+
+  it('should clear completed task', () => {
```
TimSielemann, line 65: Thx for testing!

src/main/webapp/app/tasks/default-task.service.ts
```diff
@@ -16,10 +16,18 @@ export class DefaultTaskService implements TaskService {
     return this.http.post<Task>(this.baseUrl + '/tasks', {name} as Task);
   }
 
+  update(id: string): Observable<Task> {
+    return this.http.post<Task>(this.baseUrl + '/tasks', {id} as Task);
```
TimSielemann, line 20: why is the update implemented as post? I guess this should be a put.

src/main/webapp/app/tasks/default-task.service.ts
```diff
@@ -16,10 +16,18 @@ export class DefaultTaskService implements TaskService {
...
+  }
+
   delete(id: string): Observable<void> {
     return this.http.delete<void>(this.baseUrl + '/tasks/' + id);
   }
 
+  clearCompletedTasks(): Observable<Task[]> {
+    return this.http.get<Task[]>(this.baseUrl + '/tasks');
```
TimSielemann, line 28: Why is the clear completed tasks implemented as get, i guess this should be a delete.

src/main/webapp/app/tasks/local-task.service.spec.ts
```diff
@@ -52,4 +53,21 @@ describe('LocalTaskService', () => {
...
     expect(localStorage.setItem).toHaveBeenCalled();
   });
+
+  it('should update the status of the task', () => {
+    const data = {id: 'db1c4afe-d358-4b14-acfd-0953e882a6b6', name: 'Test Task', isCompleted: false};
+    // when
+    taskService.update(id, data);
+
+    // then
+    expect(data).toEqual({id: 'db1c4afe-d358-4b14-acfd-0953e882a6b6', name: 'Test Task', isCompleted: true});
```
TimSielemann, line 63: you should rather test the return of the service method

src/main/webapp/app/tasks/local-task.service.spec.ts
```diff
@@ -52,4 +53,21 @@ describe('LocalTaskService', () => {
...
+  });
+
+  it('should update the local storage with pending tasks', () => {
+    // when
+    taskService.clearCompletedTasks(mockTask);
+
+    // then
+    expect(localStorage.setItem).toHaveBeenCalled();
```
TimSielemann, line 71: you can test how the setItem is called with the method toHaveBeenCalledWith(). This would secure that you are deleting all tasks that completet

src/main/webapp/app/tasks/local-task.service.spec.ts
```diff
@@ -52,4 +53,21 @@ describe('LocalTaskService', () => {
...
+    // when
+    taskService.update(id, data);
+
+    // then
+    expect(data).toEqual({id: 'db1c4afe-d358-4b14-acfd-0953e882a6b6', name: 'Test Task', isCompleted: true});
+  });
+
+  it('should update the local storage with pending tasks', () => {
+    // when
+    taskService.clearCompletedTasks(mockTask);
```
TimSielemann, line 68: this shouldn't compile because you need to use give it an array but it does because you did not give the parameter a type. Additionally it would be good if you give it at least one completed and one uncompleted tasks to test the functionality.

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -16,25 +16,43 @@ export class LocalTaskService implements TaskService {
...
     const tasks = this.readTasks();
-    const task = {id: uuid(), name};
+    // Added new key value pair
+    const task = {id: uuid(), name, isCompleted: false};
     tasks.push(task);
     this.writeTasks(tasks);
     return of(task);
   }
 
+  update(id: string, task: Task): Observable<Task> {
```
TimSielemann, line 26: Why did you decide to do the change by mutating the task?

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -16,25 +16,43 @@ export class LocalTaskService implements TaskService {
...
+    const tasks = this.readTasks();
+    const index = tasks.findIndex(t => t.id === id);
+
+    if (index !== -1) {
+      task.isCompleted = !task.isCompleted;
```
TimSielemann, line 31: Feels a bit odd to change the parameter value

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -16,25 +16,43 @@ export class LocalTaskService implements TaskService {
...
+    const index = tasks.findIndex(t => t.id === id);
     if (index !== -1) {
       tasks.splice(index, 1);
       this.writeTasks(tasks);
     }
     return of(null);
   }
 
+  clearCompletedTasks(tasks): void {
+    this.writeTasks(tasks);
```
TimSielemann, line 50: As said would be better to give the tasks parameter a type. 

I would assume some kind of filtering here would be better than doing it in the component and simply save the given tasks here

src/main/webapp/app/tasks/task-list/task-list.component.scss
```diff
@@ -1,14 +1,35 @@
+$brand_color: #9bbf29;
```
TimSielemann, line 1: Nice :)

src/main/webapp/app/tasks/task-list/task-list.component.spec.ts
```diff
@@ -48,9 +49,37 @@ describe('TaskListComponent', () => {
...
-    expect(deleteEmitter).toHaveBeenCalledWith({id: 'id', name: 'My task'});
+    expect(deleteEmitter).toHaveBeenCalledWith({id: 'id', name: 'My task', isCompleted: false});
+  });
+
+  it('should update the task', () => {
+    taskService.update.and.returnValue(of(null));
+
+    component.update({id: 'id', name: 'My task', isCompleted: true});
+
+    expect(taskService.update).toHaveBeenCalled();
```
TimSielemann, line 63: Again toHaveBeenCalledWith({...task}) would be better solution here

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -20,9 +23,38 @@ export class TaskListComponent {
 
   constructor(@Inject('TaskService') private taskService: TaskService) { }
 
+  update(task: Task): void {
+    this.taskService.update(task.id, task)
+      // This is a great method to unsubscribe from observable when it destoyed
+      .pipe(untilDestroyed(this))
```
TimSielemann, line 29: yeah this is nice

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -20,9 +23,38 @@ export class TaskListComponent {
...
+      .subscribe(() => {
+        this.deleted.emit(task);
     });
   }
+
+  clearCompletedTasks(tasks: Task[]): void {
+    // Optimistic approach
+    const currentTasks = tasks.filter( task => task.isCompleted === false );
```
TimSielemann, line 46: As said if i name the service method "clearCompletedTasks" i would assume that the service is taking care of filtering out all completed tasks. Another solution would be to filter for completed tasks and add a service method "deleteAll" or something. I would prefer the first solution.

package.json
```diff
@@ -23,6 +23,7 @@
     "@angular/platform-browser": "11.2.5",
     "@angular/platform-browser-dynamic": "11.2.5",
     "@angular/router": "11.2.5",
+    "@ngneat/until-destroy": "^8.0.4",
```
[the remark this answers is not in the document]
shafeequeonline: Thanks 😄

src/main/webapp/app/shared/components/checkbox/checkbox.component.html
shafeequeonline: Great catch regarding the click handler. We don't need this at this point. 

Regarding the mat-checkbox I had some plans to make this a global component by adding few more options, then this can be used by other areas also. Using shared components will give us the freedom to update to any libraries, for example, if we use PrimeNG later we have to change in one place.

src/main/webapp/app/shared/components/checkbox/checkbox.component.ts
shafeequeonline: Good catch 👍 

In this context, it's not useful. But I wanted to demonstrate that we can push changes or listen to the events from a shared component. As an example, if we have a shared textbox component we can update the value from the input box. Then we can listen to that on the onChange method with $event.target.value. These steps to be added to the shared `(click)="pushChanges('')"` in previous comment.

src/main/webapp/app/shared/shared.module.ts
shafeequeonline: Thanks 👍

src/main/webapp/app/tasks/default-task.service.spec.ts
shafeequeonline: Thank you 👍

src/main/webapp/app/tasks/default-task.service.ts
shafeequeonline: yes, you are right. Thanks for the feedback 👍

src/main/webapp/app/tasks/local-task.service.spec.ts
shafeequeonline: Thank you 👍

src/main/webapp/app/tasks/local-task.service.spec.ts
shafeequeonline: Thanks, great input.

src/main/webapp/app/tasks/local-task.service.spec.ts
shafeequeonline: Great. That's nice logic, Thanks. I really appreciate this.

src/main/webapp/app/tasks/local-task.service.ts
shafeequeonline: I am updating the task by the status of isCompleted inside the array. That's why I have used this method.

src/main/webapp/app/tasks/local-task.service.ts
shafeequeonline: Oh my bad, I missed mentioning the type here. 

Regarding the second point, good input. Thanks. You are right. this can be done at the component level.

src/main/webapp/app/tasks/task-list/task-list.component.scss
shafeequeonline: Thanks 😄 . It is a good practice to make any updates when we identify an improvement while we are working on the same file.

src/main/webapp/app/tasks/task-list/task-list.component.spec.ts
shafeequeonline: Thanks 👍 . This is really valuable input.

src/main/webapp/app/tasks/task-list/task-list.component.ts
shafeequeonline: Thank you 😄

src/main/webapp/app/tasks/task-list/task-list.component.ts
shafeequeonline: great input 👍 , Thanks

PR #122, opened 2021-05-21
A user can mark task as done

Beatriz-Porto:


pixartist (commented):
- None of the optional tasks was worked on
- Done tasks can be hidden, but not removed
- No actual implementation was written in the local or default task-service, no changes are persisted
- No tests were written

PR #123, opened 2021-05-28
A user can search for tasks

npahal:


src/main/webapp/app/tasks/search.pipe.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (searchQuery.indexOf(' AND ') !== -1 && searchQuery.indexOf(' OR ') !== -1) {
+      /**
+       * 1. split searchQuery using ' OR ' to get optional queries
+       * 2. if searchQuery does not contain any OR conditions then 1 sub-query will have all AND conditions
+       * 2. iterate through sub-queries and check if sub-queries have any more conditions, if yes then evaluate using sub-query
+       * 3. if no, then concat search results and return final result
+       */
+      const querySets = searchQuery.split(' OR ');
+      return querySets.reduce((results: Task[], query) => {
+        if (query.indexOf(' AND ') !== -1 || query.indexOf(' OR ') !== -1) {
```
fynnfeldpausch, line 19: the 2nd part of the condition will never be true.

PR #124, opened 2021-05-30
A user can change task status - #29

hsallabas:
Description: Please test with the local environment.

Feature: 
- Change task status ("To Do" and "Done" are default status)
- Change task order
- Clear all done tasks
- Add new status ("In Progress", "Blocked", ...)
- Delete status
- Set status order (using drag and drop)
- Drag and Drop support
- Test code

test-results

screenshot-1

screenshot-2

screenshot-3

### DIFF
--- a/src/main/webapp/app/shared/components/modal/modal.component.spec.ts
+++ b/src/main/webapp/app/shared/components/modal/modal.component.spec.ts
@@ -1,0 +1,25 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { ModalComponent } from './modal.component';
+
+describe('ModalComponent', () => {
+  let component: ModalComponent;
+  let fixture: ComponentFixture<ModalComponent>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      declarations: [ ModalComponent ]
+    })
+    .compileComponents();
+  });
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(ModalComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
--- a/src/main/webapp/app/shared/components/modal/modal.component.spec.ts
+++ b/src/main/webapp/app/shared/components/modal/modal.component.spec.ts
@@ -1,14 +1,26 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { MatDialogModule, MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { StatusFormComponent } from 'app/tasks/status-form/status-form.component';
 
 import { ModalComponent } from './modal.component';
 
 describe('ModalComponent', () => {
   let component: ModalComponent;
   let fixture: ComponentFixture<ModalComponent>;
+  let statusFormComponent: StatusFormComponent;
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
-      declarations: [ ModalComponent ]
+      declarations: [ ModalComponent ],
+      providers: [{
+        provide: MatDialogRef,
+        useValue: {}
+      },
+      { provide: MAT_DIALOG_DATA, useValue: { component: statusFormComponent } }
+    ],
+      imports: [
+        MatDialogModule
+      ]
     })
     .compileComponents();
   });
--- a/src/main/webapp/app/tasks/status-form/status-form.component.html
+++ b/src/main/webapp/app/tasks/status-form/status-form.component.html
@@ -1,0 +1,36 @@
+<div class="mat-typography">
+    <div class="wrapper" *ngIf="tasks$ | async as tasks">
+        <div class="status-form-wraper">
+            <form [formGroup]="statusForm" (ngSubmit)="onSubmit(tasks)">
+                <mat-form-field>
+                    <input type="text" formControlName="name" matInput placeholder="Add a new status"
+                        data-cy="task-input">
+                    <mat-icon matSuffix aria-label="Add Status" (click)="onSubmit(tasks)">send</mat-icon>
+                </mat-form-field>
+            </form>
+        </div>
+        <div class="drag-and-drop-title">
+            <p>Drag & drop to sort</p>
+        </div>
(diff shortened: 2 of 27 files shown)


src/main/webapp/app/tasks/status-form/status-form.component.html
```diff
@@ -0,0 +1,36 @@
...
+        </div>
+        <div 
+            cdkDropList
+            [cdkDropListData]="tasks"
+            (cdkDropListDropped)="drop($event, tasks)"
+            [cdkDropListSortPredicate]="preventMainStatus" 
+            class="example-list">
+            <div 
+                class="example-box" 
+                *ngFor="let task of tasks; trackBy: trackByFn; let i = index"
```
TimSielemann, line 23: shouldn't those variables be called "status" 😆

src/main/webapp/app/tasks/status-form/status-form.component.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  beforeEach(() => {
+    fixture = TestBed.createComponent(StatusFormComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
```
TimSielemann, line 32: tests missing

src/main/webapp/app/tasks/status-form/status-form.component.ts
```diff
@@ -0,0 +1,89 @@
...
+
+@Component({
+  selector: 'tiny-status-form',
+  templateUrl: './status-form.component.html',
+  styleUrls: ['./status-form.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class StatusFormComponent implements OnInit {
+
+  tasks$: Observable<TaskList[] | Task[]>;
```
TimSielemann, line 20: this should be mapped to a string array of status names in the component to prevent business logic in the template

src/main/webapp/app/tasks/status-form/status-form.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  ngOnInit(): void {
+    this.tasks$ = this.taskService.getAll();
+  }
+
+  /**
+   * drag and drop event
+   * @param event CdkDragDrop
+   * @param taskList list of all tasks
+   */
+  drop(event: CdkDragDrop<string[]>, taskList: TaskList[]): void {
```
TimSielemann, line 38: i would see this functionality in the service tbh like "reorderStatus" or something (same goes for all taskList mutating functionality in this component).

src/main/webapp/app/tasks/status-form/status-form.component.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.tasks$ = this.taskService.updateAll(tasksList);
+  }
+
+  /**
+   * TrackBy to increase performance
+   * @param index item index number
+   * @returns item index number
+   */
+  trackByFn(index: number): number {
+    return index;
```
TimSielemann, line 86: isn't a trackByFunction by index the wrong way for a list where the index changes because of the drag and drop. I guess you are lucky here because you are not using the async pipe in the ngFor.

src/main/webapp/app/tasks/task-form/task-form.component.ts
```diff
@@ -13,18 +15,28 @@ import { TaskService } from '../task.service';
...
+
   taskForm: FormGroup = new FormGroup({
     name: new FormControl('', Validators.required)
   });
 
   constructor(@Inject('TaskService') private taskService: TaskService) { }
 
+  public ngOnDestroy(): void {
+    this.unsubscribe$.next(true);
+    this.unsubscribe$.unsubscribe();
```
TimSielemann, line 32: shouldn't it be "complete"?

src/main/webapp/app/shared/components/modal/modal.component.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { MatDialogModule, MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { StatusFormComponent } from 'app/tasks/status-form/status-form.component';
+
+import { ModalComponent } from './modal.component';
+
+describe('ModalComponent', () => {
+  let component: ModalComponent;
+  let fixture: ComponentFixture<ModalComponent>;
+  let statusFormComponent: StatusFormComponent;
```
TimSielemann, line 10: you never assign this to anything

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -28,5 +42,5 @@ export interface TaskService {
    * @param id the ID of the task to be removed
    * @returns an empty `Observable`
    */
-  delete(id: string): Observable<void>;
+  delete(id: string, statusIndex?: number): Observable<void>;
```
TimSielemann, line 45: Status index shouldn't be optional as you will get an error when calling the actual function in the local task service (an other way would be to implement the service function handling that)

src/main/webapp/app/tasks/status-form/status-form.component.spec.ts
hsallabas: I thought I had exceeded the allotted time for the task too much so I couldn't develop it. Besides, I need to improve my test skills.

src/main/webapp/app/tasks/status-form/status-form.component.ts
hsallabas: It is a copy-paste issue. I could change it to reorderStatus. I didn't think naming will be a problem in the interview. I have focused on logic.

src/main/webapp/app/tasks/status-form/status-form.component.ts
hsallabas: :)

src/main/webapp/app/tasks/task.service.ts
hsallabas: I got an error from default-task-service.ts I didn't want to change or delete it so I preferred to do optional. I didn't think another way. I could solve it as you say.

PR #126, opened 2021-06-09
Implementation of 'A user can search for tasks'

lschmischke:
Implemented ['A user can search for tasks'](https://github.com/mindsmash/tiny-tasks/issues/28).

Implemented functionality:

- The user can search tasks via a search input
- The search is performaned instantaneously
- The search can be reset with a single click
  - A button to clear the field appears if the field is not empty
- The user can deep-link to a search result (e.g. call URL?q=search and find the search pre-filled). 
  - To help with sharing of the result, the URL will automatically update itself with the query parameter when the search field is used.
- The user can use search operators. 
  - AND and OR are implemented, multiple operators can be chained, however AND has precedence and grouping of expressions via parantheses is not supported.

Looking forward to any comments.


src/main/webapp/app/tasks/task-form/task-form.component.spec.ts
```diff
@@ -16,7 +18,11 @@ describe('TaskFormComponent', () => {
       providers: [{
         provide: 'TaskService',
         useValue: taskService
-      }]
+      }],
+      imports: [
```
ddoe8, line 22: Why are you importing those modules for this test?

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,5 +1,8 @@
+
+<tiny-task-search #search></tiny-task-search>
```
ddoe8, line 2: the search component should have been added to the app.component and the logic programed to make the search/filter work that way. Also careful with adding extra empty line

src/main/webapp/app/tasks/tasks.module.ts
```diff
@@ -1,24 +1,31 @@
...
-import { ReactiveFormsModule } from '@angular/forms';
+import {FormsModule, ReactiveFormsModule} from '@angular/forms';
 import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
 import { MatInputModule } from '@angular/material/input';
 import { MatListModule } from '@angular/material/list';
 
 import { TaskFormComponent } from './task-form/task-form.component';
 import { TaskListComponent } from './task-list/task-list.component';
+import { TaskSearchFilterPipe } from './task-search/task-search-filter.pipe';
```
ddoe8, line 11: you can organize the imports to keep a consistent style

PR #140, opened 2021-08-16
#29 a user can mark a task as done

isakaliu:
- a user can clear done tasks
- a sorting is also added
- a user can delete the task
- local-config used

### DIFF
--- a/src/main/webapp/app/tasks/default-task.service.ts
+++ b/src/main/webapp/app/tasks/default-task.service.ts
@@ -23,4 +23,12 @@ export class DefaultTaskService implements TaskService {
   getAll(): Observable<Task[]> {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
   }
+  editTask(task: Task): any {
+  }
+
+  sortTasks(): any {
+  }
+
+  clearDoneTasks(): any {
+  }
 }
--- a/src/main/webapp/app/tasks/local-task.service.ts
+++ b/src/main/webapp/app/tasks/local-task.service.ts
@@ -14,9 +14,9 @@ export class LocalTaskService implements TaskService {
     return of(this.readTasks());
   }
 
-  create(name: string): Observable<Task> {
+  create(name: string, status: string): Observable<Task> {
     const tasks = this.readTasks();
-    const task = {id: uuid(), name};
+    const task = {id: uuid(), name, status};
     tasks.push(task);
     this.writeTasks(tasks);
     return of(task);
@@ -40,4 +40,34 @@ export class LocalTaskService implements TaskService {
   private writeTasks(tasks: Task[]): void {
     localStorage.setItem(LocalTaskService.STORAGE_KEY, JSON.stringify(tasks));
   }
+  editTask(task: Task): any {
+    const tasks = this.readTasks();
+    const item = tasks.find(x => x.id === task.id);
+    if (item){
+      item.status = 'done';
+    }
+    localStorage.setItem(LocalTaskService.STORAGE_KEY, JSON.stringify(tasks));
+    this.sortTasks();
+  }
+
+  sortTasks(): any{
+    const tasks =  this.readTasks();
+    tasks.sort((a, b) => {
+      if (a.status === 'done') { return 1; }
+      else  if (a.status === 'uncompleted'){
+        return -1;
+      }else{
+        return 0;
+      }
+    });
+    localStorage.setItem(LocalTaskService.STORAGE_KEY, JSON.stringify(tasks));
+  }
+
+  clearDoneTasks(): any{
+    const items = this.readTasks();
+    const uncompleted = items.filter(x => x.status === 'uncompleted');
+    localStorage.setItem(LocalTaskService.STORAGE_KEY, JSON.stringify(uncompleted));
+
+  }
+
 }
--- a/src/main/webapp/app/tasks/task-list/task-list.component.html
+++ b/src/main/webapp/app/tasks/task-list/task-list.component.html
@@ -1,8 +1,23 @@
+<button mat-raised-button id="basic" *ngIf="this.tasks.length > 0"  style=" margin:20px" color="primary" type="button" mat-button
+        (click)="editTasks()">
+  <span *ngIf="!editMode">Edit Task</span>
+  <span *ngIf="editMode">Done</span>
+</button>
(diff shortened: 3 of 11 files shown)


src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -12,17 +22,77 @@ import { TaskService } from '../task.service';
...
+   */
+  editTasks(): void{
+    this.editMode = !this.editMode;
+  }
+
+  /**
+   * check the selected task if true modify css class to be selected
+   * @param task
+   */
+  checkSelectedTask(task: Task): any{
```
fynnfeldpausch, line 59: 1. Don't use any as a return type but use proper typings
2. What is the difference to the implementation of `getStatus`
3. avoid the if conditions (in both implementations) and return go for something like `return !!item`

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -12,17 +22,77 @@ import { TaskService } from '../task.service';
...
+  changeTaskStatus(task: Task): void{
+    this.taskService.editTask(task);
+    this.updateTasks();
+    this.taskService.sortTasks();
+
+  }
+
+
+  getStatus(task: Task): any{
+    const item = this.tasks.find(x => x.id === task.id &&  x.status === 'done')
```
fynnfeldpausch, line 80: See comments for `checkSelectedTask`

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,8 +1,23 @@
+<button mat-raised-button id="basic" *ngIf="this.tasks.length > 0"  style=" margin:20px" color="primary" type="button" mat-button
+        (click)="editTasks()">
+  <span *ngIf="!editMode">Edit Task</span>
+  <span *ngIf="editMode">Done</span>
+</button>
+<button mat-raised-button id="basic" *ngIf="this.tasks.length > 0"  style=" margin:20px" color="primary" type="button" mat-button
```
fynnfeldpausch, line 6: avoid inline styles

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -1,8 +1,23 @@
...
```
fynnfeldpausch, line 1: avoid inline styles

src/main/webapp/app/tasks/default-task.service.ts
```diff
@@ -23,4 +23,12 @@ export class DefaultTaskService implements TaskService {
   getAll(): Observable<Task[]> {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
   }
+  editTask(task: Task): any {
```
fynnfeldpausch, line 26: use proper typings instead of any

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -14,9 +14,9 @@ export class LocalTaskService implements TaskService {
     return of(this.readTasks());
   }
 
-  create(name: string): Observable<Task> {
+  create(name: string, status: string): Observable<Task> {
```
fynnfeldpausch, line 17: Status should be an enum and not a string

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -40,4 +40,34 @@ export class LocalTaskService implements TaskService {
   private writeTasks(tasks: Task[]): void {
     localStorage.setItem(LocalTaskService.STORAGE_KEY, JSON.stringify(tasks));
   }
+  editTask(task: Task): any {
```
fynnfeldpausch, line 43: editTask is very misleading name in this case

fynnfeldpausch: use proper typings instead of any

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -40,4 +40,34 @@ export class LocalTaskService implements TaskService {
...
+    const tasks = this.readTasks();
+    const item = tasks.find(x => x.id === task.id);
+    if (item){
+      item.status = 'done';
+    }
+    localStorage.setItem(LocalTaskService.STORAGE_KEY, JSON.stringify(tasks));
+    this.sortTasks();
```
fynnfeldpausch, line 50: why not sort them before storing instead of storing the items, then reading them again, sorting and storing yet again?

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -40,4 +40,34 @@ export class LocalTaskService implements TaskService {
...
+  }
+
+  sortTasks(): any{
+    const tasks =  this.readTasks();
+    tasks.sort((a, b) => {
+      if (a.status === 'done') { return 1; }
+      else  if (a.status === 'uncompleted'){
+        return -1;
+      }else{
+        return 0;
```
fynnfeldpausch, line 60: what are other statuses?

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -40,4 +40,34 @@ export class LocalTaskService implements TaskService {
...
+      }
+    });
+    localStorage.setItem(LocalTaskService.STORAGE_KEY, JSON.stringify(tasks));
+  }
+
+  clearDoneTasks(): any{
```
fynnfeldpausch, line 66: this method is not tested

src/main/webapp/app/tasks/task-list/task-list.component.scss
```diff
@@ -12,3 +12,7 @@
     color: #101d30;
   }
 }
+
+.selectedTask{
+  background-color: grey;;
```
fynnfeldpausch, line 17: not a big issue but just unclean: `;;`

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -29,4 +30,21 @@ export interface TaskService {
    * @returns an empty `Observable`
    */
   delete(id: string): Observable<void>;
+
+  /**
+   * Edit the task status done or uncompleted
+   * @param task of type Task interface
+   */
+  editTask(task: Task): any;
```
fynnfeldpausch, line 38: Please use proper typings

PR #142, opened 2021-08-21
feature: A user can mark a task as done #29

alefiori:
Implemented possibility to mark a task as done in order to solve #29.

### DIFF
--- a/src/main/webapp/app/tasks/local-task.service.ts
+++ b/src/main/webapp/app/tasks/local-task.service.ts
@@ -34,9 +34,9 @@ export class LocalTaskService implements TaskService {
 
   setDone(id: string): Observable<void> {
     const tasks = this.readTasks();
-    const task = tasks.find(task => task.id === id);
-    if (task) {
-      task.done = !task.done;
+    const selectedTask = tasks.find(task => task.id === id);
+    if (selectedTask) {
+      selectedTask.done = !selectedTask.done;
       this.writeTasks(tasks);
     }
     return of(null);
@@ -48,6 +48,7 @@ export class LocalTaskService implements TaskService {
   }
 
   private writeTasks(tasks: Task[]): void {
+    tasks.sort(task => task.done ? 1 : -1);
     localStorage.setItem(LocalTaskService.STORAGE_KEY, JSON.stringify(tasks));
   }
 }
--- a/src/main/webapp/app/tasks/local-task.service.ts
+++ b/src/main/webapp/app/tasks/local-task.service.ts
@@ -32,6 +32,16 @@ export class LocalTaskService implements TaskService {
     return of(null);
   }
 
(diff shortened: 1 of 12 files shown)


src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -32,12 +32,30 @@ export class LocalTaskService implements TaskService {
     return of(null);
   }
 
+  setDone(id: string): Observable<void> {
```
fynnfeldpausch, line 35: just a minor remark, but if you actually go for a toggle implementation here (`selectedTask.done = !selectedTask.done`), I'd tend more towards a name that reflects that. So either call it "toggleDone" or maybe make it even more explicit and add a boolean attribute to the method's signature.

src/main/webapp/app/tasks/task-list/task-list.component.html
```diff
@@ -2,8 +2,15 @@
   <mat-list-item *ngFor="let task of tasks" class="mat-elevation-z1">
     <mat-icon mat-list-icon>assignment</mat-icon>
     <h4 mat-line>{{task.name}}</h4>
-    <button mat-icon-button aria-label="Delete task" color="primary">
+    <button [ngClass]="task.done ? 'done-button--selected':'done-button'" mat-icon-button aria-label="Task done"
```
fynnfeldpausch, line 5: If the button is implemented to reflect a toggle value, one could improve the accessibility even more by providing additional ARIA attributes (e.g. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role)

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -29,4 +29,8 @@ export interface TaskService {
    * @returns an empty `Observable`
    */
   delete(id: string): Observable<void>;
+
+  setDone(id: string): Observable<void>;
```
fynnfeldpausch, line 33: missing docs here

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -29,4 +29,8 @@ export interface TaskService {
...
+
+  emptyDoneList(): Observable<void>;
```
fynnfeldpausch, line 35: missing docs here

src/main/webapp/app/tasks/task.ts
```diff
@@ -4,4 +4,5 @@
 export interface Task {
   id: string;
   name: string;
+  done?: boolean;
```
fynnfeldpausch, line 7: I'd not go for an optional property here as it introduces some weird ternary logic on the done state

fynnfeldpausch (approved):
Nice and clean solution. A few minor remarks, but nothing serious.

PR #143, opened 2021-08-25
#29 Change task status

NiushaGhanbari:
Ticket => https://github.com/mindsmash/tiny-tasks/issues/29

### DIFF
--- a/src/main/webapp/app/tasks/default-task.service.ts
+++ b/src/main/webapp/app/tasks/default-task.service.ts
@@ -3,7 +3,7 @@ import { Inject, Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
 
 import { BASE_URL } from '../app.tokens';
-import { Task } from './task';
+import { Task, TaskStatus } from './task';
 import { TaskService } from './task.service';
 
 @Injectable()
@@ -22,5 +22,12 @@ export class DefaultTaskService implements TaskService {
 
   getAll(): Observable<Task[]> {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
+  }
+
+  updateStatus(id: string, status: TaskStatus): Observable<Task> {
+    return this.http.put<Task>(this.baseUrl + '/tasks/' + id + '/status', { status });
+  }
+  clearDoneTasks(): Observable<Task> {
+    return this.http.put<Task>(this.baseUrl + '/tasks/clear', { status });
   }
 }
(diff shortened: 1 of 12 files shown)


src/main/webapp/app/tasks/default-task.service.ts
```diff
@@ -23,4 +23,11 @@ export class DefaultTaskService implements TaskService {
   getAll(): Observable<Task[]> {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
   }
+
+  updateStatus(id: string, status: TaskStatus): Observable<Task> {
+    return this.http.put<Task>(this.baseUrl + '/tasks/' + id + '/status', { status });
+  }
+  clearDoneTasks(): Observable<Task> {
+    return this.http.put<Task>(this.baseUrl + '/tasks/clear', { status });
```
fynnfeldpausch, line 31: This would probably be a DELETE call.

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -29,4 +29,18 @@ export interface TaskService {
    * @returns an empty `Observable`
    */
   delete(id: string): Observable<void>;
+
+
+  /**
+   *
```
fynnfeldpausch, line 35: Incomplete docs

src/main/webapp/app/tasks/task.service.ts
```diff
@@ -29,4 +29,18 @@ export interface TaskService {
...
+   * @param id the ID of the task to be status changed
+   * @param status new task status
+   * @returns an `Observable` holding the task with new status
+   */
+  updateStatus(id: string, status: TaskStatus): Observable<Task>;
+   /**
+    *
```
fynnfeldpausch, line 42: Incomplete docs

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -17,12 +18,45 @@ export class TaskListComponent {
...
     });
   }
+
+  changeStatus(task: Task, status: TaskStatus): void {
+    this.taskService.updateStatus(task.id, status).subscribe((updatedTask: Task) => {
+      this.statusChanged.emit(updatedTask);
+    });
+  }
+
+  showOption(task: Task, status: TaskStatus): boolean {
```
fynnfeldpausch, line 41: There are a couple of tests missing for these methods

fynnfeldpausch (commented):
Solid implementation, thank you! Just a few minor remarks.

src/main/webapp/app/tasks/default-task.service.ts
niusha1993: Nice finding! Yes, you are right, I knew it but forgot it. so a `DELETE` HTTP verb should be used when we are deleting a resource 👍

src/main/webapp/app/tasks/task.service.ts
niusha1993: Could you give me a hint about it?

PR #145, opened 2021-08-30
feat: new feature to mark task as completed

vigneswaran-js-dev:


vigneswaran-js-dev:
Screenshot 2021-08-30 at 12 49 09 PM
Screenshot 2021-08-30 at 12 48 41 PM
Screenshot 2021-08-30 at 12 51 10 PM

ddoe8:
Can you provide a short summary on "What has been done"?

src/main/webapp/app/app.component.ts
```diff
@@ -28,4 +28,14 @@ export class AppComponent implements OnInit {
   deleted(): void {
     this.tasks$ = this.taskService.getAll();
   }
+
+  completed(): void {
+    this.tasks$ = this.taskService.getAll();
+  }
+
+  async clearCompleted(): Promise<void> {
```
ddoe8, line 36: Why the change to use Promises over continuing using Observable?

src/main/webapp/app/tasks/default-task.service.ts
```diff
@@ -23,4 +23,12 @@ export class DefaultTaskService implements TaskService {
   getAll(): Observable<Task[]> {
     return this.http.get<Task[]>(this.baseUrl + '/tasks');
   }
+
+  updateStatus(id: string, status: boolean): Observable<void> {
+    return this.http.patch<void>(`${this.baseUrl}/tasks/${id}`, { completed: status });
```
ddoe8, line 28: Why the use of patch over put (for example)?

src/main/webapp/app/tasks/default-task.service.spec.ts
```diff
@@ -61,4 +61,29 @@ describe('DefaultTaskService', () => {
...
+    const req = httpTestingController.expectOne(request => request.url === 'http://backend.tld/tasks/id123');
+    expect(req.request.method).toEqual('PATCH');
+
+    // finally
+    req.flush({});
+  });
+
+  it('should clear completed tasks', () => {
+    // when
+    const tasks$ = [{id: 'id123', name: 'Do something', completed: false}];
```
ddoe8, line 79: when using the "given when then" approach on unit testing the task should be on given

src/main/webapp/app/tasks/default-task.service.ts
vigneswaran-js-dev: When ever we are replacing the resource entirely, we can use PUT. When we doing a partial update, we can use HTTP PATCH. Here in this scenario we are jus updating completed status of a task. So PATCH is preferable.

src/main/webapp/app/tasks/default-task.service.spec.ts
vigneswaran-js-dev: agreed. it was a miss from my end.

src/main/webapp/app/app.component.ts
vigneswaran-js-dev: To avoid callback hell and we need to make calls synchronously one after another.  

OfCourse we could have used flatMap or switchMap Rxjs operator to achieve the same behaviour but you don’t always need to use Observables. Promises are still here, and it’s OK to use them when you need it.

vigneswaran-js-dev:
> Can you provide a short summary on "What has been done"?

Added a feature for the user to mark the task to completed state.

vigneswaran-js-dev:
> Can you provide a short summary on "What has been done"?

Adding the capability for marking a task as done and ability to clear all the completed  tasks on click of a button.

Also when ever we mark a task as completed that task is pushed down to the bottom of the list and completed styles get applied to the task.

“Clear completed tasks” button is placed in app component level instead of having it in task list component for better usability and it can be used to scale for app level features.

PR #149, opened 2021-09-28
Due date and search

seaneesrehman:
A user can set a due date #27 (Front End)

- Due added with two options popup and inline

A user can search for tasks #28 (Front End)

- Search functionality has been added (Not implemented optional success cases)

### DIFF
--- a/src/main/webapp/app/Utility/common.ts
+++ b/src/main/webapp/app/Utility/common.ts
@@ -1,0 +1,10 @@
+
+export class Common {
+    static setMinDate() {
+        const now = new Date();
+        var minDate = new Date();
+        minDate.setDate(now.getDate() - 1);
+        return minDate
+      }
+    
+}
--- a/src/main/webapp/app/Utility/common.ts
+++ b/src/main/webapp/app/Utility/common.ts
@@ -1,9 +1,8 @@
 
 export class Common {
     static setMinDate() {
-        const now = new Date();
         var minDate = new Date();
-        minDate.setDate(now.getDate() - 1);
+        minDate.setDate(minDate.getDate() - 1);
         return minDate
       }
     
--- a/src/main/webapp/app/app.component.html
+++ b/src/main/webapp/app/app.component.html
@@ -8,8 +8,14 @@
     </div>
   </mat-toolbar>
   <section class="content">
+    
     <div class="wrapper">
-      <tiny-task-list [tasks]="tasks$ | async" (deleted)="deleted()"></tiny-task-list>
+      <div class="searchTest">
+      <tiny-task-search (searchDataEmitter)="receiveMessage($event)">
+      
+      </tiny-task-search>
+      </div>
+      <tiny-task-list [tasks]="tasks$ | async" (deleted)="deleted()" (updated)="updated()"></tiny-task-list>
     </div>
   </section>
 </div>
--- a/src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.html
+++ b/src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.html
@@ -1,0 +1,16 @@
+
+<mat-form-field>
+    <mat-label>Choose a Date</mat-label>
+    <input matInput [ngxMatDatetimePicker]="picker" p
+      [min]="minDate"  [disabled]="disabled" [formControl]="dateControl">
+    <mat-datepicker-toggle matSuffix [for]="$any(picker)"></mat-datepicker-toggle>
+    <ngx-mat-datetime-picker #picker [showSpinners]="showSpinners" [showSeconds]="showSeconds"
+      [stepHour]="stepHour" [stepMinute]="stepMinute" [stepSecond]="stepSecond" [touchUi]="touchUi"
+      [color]="color" [enableMeridian]="enableMeridian" [disabled]="disabledDateTime">
+    </ngx-mat-datetime-picker>
+  </mat-form-field>
+
+  <mat-dialog-actions align="end">
+    <button mat-button mat-dialog-close>Cancel</button>
+    <button mat-button [mat-dialog-close]="true" (click)="dateSelected()" cdkFocusInitial>Select</button>
+  </mat-dialog-actions>
--- a/src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
+++ b/src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
@@ -30,14 +30,10 @@ export class TaskDueDateDialogComponent {
 
   constructor(private dialogRef: MatDialogRef<TaskDueDateDialogComponent>,
     @Inject(MAT_DIALOG_DATA) public dialogData: any) {
-    this.minDate = Common.setMinDate()
+      console.log(this.dialogData)
+    this.minDate = this.dialogData
   }
 
-  private _setMinDate() {
-    const now = new Date();
-    this.minDate = new Date();
-    this.minDate.setDate(now.getDate() - 1);
-  }
 /**
  * dateSlected methos is called when user selects date from calandar
  */
--- a/src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
+++ b/src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
@@ -1,0 +1,47 @@
+import {Component, EventEmitter, Inject, Output, ViewChild} from '@angular/core';
+import {FormControl} from '@angular/forms';
+import {ThemePalette} from '@angular/material/core';
+import {MatDialogRef, MAT_DIALOG_DATA} from '@angular/material/dialog';
+import { Common } from '../../../Utility/common';
+/**
+ * A dialog created for selection of date and time 
+ */
+@Component({
+  selector: 'tiny-task-due-date-dialog',
+  templateUrl: './task-due-date-dialog.component.html',
+  styleUrls: ['./task-due-date-dialog.component.scss']
+})
+export class TaskDueDateDialogComponent {
+
+  @ViewChild('picker') picker: any;
+  public disabled = true;
+  public disabledDateTime = false;
+  public showSpinners = true;
+  public showSeconds = false;
+  public touchUi = false;
+  public enableMeridian = false;
+  public minDate: Date = new Date();
+  public stepHour = 1;
+  public stepMinute = 1;
+  public stepSecond = 1;
+  public color: ThemePalette = 'primary';
+  @Output() deleted: EventEmitter<Date> = new EventEmitter();
+  public dateControl = new FormControl();
+
+  constructor(private dialogRef: MatDialogRef<TaskDueDateDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public dialogData: any) {
+    this.minDate = Common.setMinDate()
+  }
+
+  private _setMinDate() {
+    const now = new Date();
+    this.minDate = new Date();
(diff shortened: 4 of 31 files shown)


src/main/webapp/app/Utility/common.ts
```diff
@@ -0,0 +1,10 @@
+
+export class Common {
```
TimSielemann, line 2: Ok i have several questions to this part of the code 😆 

Why didn't you just export the function instead of a static class?
Why is the min date yesterday?
Why not 
`minDate.setDate(minDate.getDate() - 1);`
?

Additionally i think this method is to specific to be a static helper method. It should be implemented in the component. ( I am no Fan of static helper functions anyway)

src/main/webapp/app/app.component.html
```diff
@@ -8,8 +8,14 @@
     </div>
   </mat-toolbar>
   <section class="content">
+
```
TimSielemann, line 11: Why are you adding the new lines to the html?

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.html
```diff
@@ -0,0 +1,16 @@
+
```
TimSielemann, line 1: Why did you decide to do a dialog time picker and an inline picker?

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  public color: ThemePalette = 'primary';
+  @Output() deleted: EventEmitter<Date> = new EventEmitter();
+  public dateControl = new FormControl();
+
+  constructor(private dialogRef: MatDialogRef<TaskDueDateDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public dialogData: any) {
+    this.minDate = Common.setMinDate()
+  }
+
+  private _setMinDate() {
```
TimSielemann, line 36: I guess this can be deleted because you solved that by that static function. In my opinion it should be given in the dialog data.

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  public minDate: Date = new Date();
+  public stepHour = 1;
+  public stepMinute = 1;
+  public stepSecond = 1;
+  public color: ThemePalette = 'primary';
+  @Output() deleted: EventEmitter<Date> = new EventEmitter();
+  public dateControl = new FormControl();
+
+  constructor(private dialogRef: MatDialogRef<TaskDueDateDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public dialogData: any) {
```
TimSielemann, line 32: Why are you injecting that if it is not needed?

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
```diff
@@ -0,0 +1,47 @@
...
+ * A dialog created for selection of date and time 
+ */
+@Component({
+  selector: 'tiny-task-due-date-dialog',
+  templateUrl: './task-due-date-dialog.component.html',
+  styleUrls: ['./task-due-date-dialog.component.scss']
+})
+export class TaskDueDateDialogComponent {
+
+  @ViewChild('picker') picker: any;
```
TimSielemann, line 16: Why are you defining it even though it is not used?

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  public showSpinners = true;
+  public showSeconds = false;
+  public touchUi = false;
+  public enableMeridian = false;
+  public minDate: Date = new Date();
+  public stepHour = 1;
+  public stepMinute = 1;
+  public stepSecond = 1;
+  public color: ThemePalette = 'primary';
+  @Output() deleted: EventEmitter<Date> = new EventEmitter();
```
TimSielemann, line 28: Output for a dialog component seems odd as everything is going into the result.

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -10,34 +11,51 @@ export class LocalTaskService implements TaskService {
...
+        return a.dueDate! > b.dueDate! ? 1:-1
+      }): [];
+    }
+    return tasks ? JSON.parse(tasks).sort(function(a:Task,b:Task){
+      return a.dueDate! > b.dueDate! ? -1:1
+    }): [];
   }
 
-  private writeTasks(tasks: Task[]): void {
+   private static writeTasks(tasks: Task[]): void {
```
TimSielemann, line 58: Why did you decide to make this static?

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -10,34 +11,51 @@ export class LocalTaskService implements TaskService {
...
   }
 
-  private readTasks(): Task[] {
+  private readTasks(search?:string): Task[] {
     const tasks = localStorage.getItem(LocalTaskService.STORAGE_KEY);
-    return tasks ? JSON.parse(tasks) : [];
+    if(search && search!.length > 0){
+      return tasks ? JSON.parse(tasks).filter((task:Task)=>{
+        return task.name.toLowerCase().includes(search.toLowerCase())
+      }).sort(function(a:Task,b:Task){
```
TimSielemann, line 49: why are you not using lambda functions here

src/main/webapp/app/tasks/local-task.service.ts
```diff
@@ -10,34 +11,51 @@ export class LocalTaskService implements TaskService {
...
+        return a.dueDate! > b.dueDate! ? 1:-1
+      }): [];
+    }
+    return tasks ? JSON.parse(tasks).sort(function(a:Task,b:Task){
```
TimSielemann, line 53: you could remove the redundant sorting code by sorting tasks before filtering (or some other option). 

BTW Sorting is not working when i am starting the application.

src/main/webapp/app/tasks/task-form/task-form.component.scss
```diff
@@ -10,3 +10,8 @@ form {
 .mat-icon {
   cursor: pointer;
 }
+
+.errorDiv {
```
TimSielemann, line 14: best practice for css classes is kebab case. Additionally i can't find any usages of this class.

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -12,17 +16,63 @@ import { TaskService } from '../task.service';
...
+  }
 
   delete(task: Task): void {
     this.taskService.delete(task.id).subscribe(() => {
       this.deleted.emit(task);
     });
   }
+
+ 
+  getObject(task: Task) {
```
TimSielemann, line 53: I guess a more speaking name would be benefitial here

src/main/webapp/app/tasks/task-list/task-list.component.ts
```diff
@@ -12,17 +16,63 @@ import { TaskService } from '../task.service';
...
+      this.updated.emit(task)
+    })
+  }
+
+  /**
+ * This will open a calandar in pop up
+ */
+  openCalander(task: Task): void {
+    const dialogRef = this.dialog.open(TaskDueDateDialogComponent);
+    dialogRef.afterClosed().subscribe(result => {
```
TimSielemann, line 68: you don't need a nested subscribe here, you can map it by an rxjs operator

src/main/webapp/app/tasks/task-search/task-search.component.scss
```diff
@@ -0,0 +1,8 @@
+.mat-form-field{
+    width: 100%;
+}
+
+.mt{
+    margin-top: 1rem;
+}
+::ng-deep  .mat-focused .mat-form-field-underline .mat-form-field-ripple{ background: greenyellow }
```
TimSielemann, line 8: ng-deep is deprecated.

src/main/webapp/app/app.component.html
seaneesrehman: By new line means the new code then it is for the search feature that I have Implemented.

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.html
seaneesrehman: Personally I like inline picker , I don't know which approach you guys follow in Coyo , so I Implemented both. It could be a design decision.

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
seaneesrehman: Yes this can be deleted. 
Yes you are right this can be implemented through dialog data ,  for dialog data i have a code setup but some how missed that.

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
seaneesrehman: Yes right this should be removed

src/main/webapp/app/tasks/local-task.service.ts
seaneesrehman: This method is used to store data in local storage. So this is a practice that when we normally play with local storage we made that function static.

src/main/webapp/app/tasks/local-task.service.ts
seaneesrehman: Usually I use arrow function but in sort i missed it.  In filter arrow function is used

src/main/webapp/app/tasks/local-task.service.ts
seaneesrehman: Sorting is not working because empty cases were not handled properly. Sorting used after filtering because it reduces the time complexity.
I have fixed it

src/main/webapp/app/tasks/task-form/task-form.component.scss
seaneesrehman: I had added this class for testing purpose , It should be removed and regarding the kebab case yes it is indeed a best practice

src/main/webapp/app/tasks/task-list/task-list.component.ts
seaneesrehman: yes it should be getDueDate()

src/main/webapp/app/tasks/task-search/task-search.component.scss
seaneesrehman: Acknowledged, Will look for work around.

src/main/webapp/app/Utility/common.ts
seaneesrehman: 1 - I implemented  it as class because it has been used for any common features through out the  app if any, so in this perspective i used. (In this case two components are using this same method)
Yes you right it can be implemented through export function also. 
It's not yesterday , selection is allowed from today onward. It is library behaviour.

src/main/webapp/app/tasks/dialog/task-due-date-dialog/task-due-date-dialog.component.ts
seaneesrehman: I added it for use but forgot it during that common class implementation.

PR #154, opened 2022-10-29
Issue 27 & Issue 28

paul-luchian:
BUG28
- created test-coverage script
- created basic style framework
- created filter section
- integrated api call with the fitlers
- integrated queryParams with the filter
- created api call for getFiltered tasks
- created utils functions
- created tests
- fixed failing tests

BUG27
- implemented due date on tasks
- implemented due date SAVE on tasks
- implemented task sorting by default||name||dueDate
- implemented task sorting direction
- integrated sorting with query params
- created tests
- improved BUG28 fix
- fixed failing tests
- created more style classes in the css framework


src/main/webapp/app/shared/functions/query-params.utils.ts
```diff
@@ -0,0 +1,10 @@
+
+const buildQueryParamString: (...params: string[]) => string = (...params: string[]): string => {
+    return params.reduce((total, next) => (total + (next ? `${next}&` : '')), '').slice(0, -1);
```
fynnfeldpausch, line 3: Why are you not using standard Angular functionality with the `httpClient`'s build-in arguments?

src/main/webapp/theme/utilities/utilities.scss
```diff
@@ -0,0 +1,14 @@
+@import "./generator.scss";
```
fynnfeldpausch, line 1: Nice one, but I guess a bit overkill for such a small project. It generates a huge amount of unused CSS.

src/main/webapp/app/shared/components/filter/filter.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+@Component({
+  selector: 'tiny-filter',
+  templateUrl: './filter.component.html',
+  styleUrls: ['./filter.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FilterComponent implements OnInit, OnDestroy {
+
+  @Output() formChanged: EventEmitter<FormGroup> = new EventEmitter<FormGroup>();
```
fynnfeldpausch, line 15: I would call this an anti-pattern to expose the entire form here. Why not implement a custom `ControlValueAccessor` here to encapsulate the behaviour?

src/main/webapp/theme/utilities/utilities.scss
paul-luchian: I usually like to do things the right way from the beginning. Like having global classes defined to reduce code duplicate/ different naming and so on. That's why I created those functions that are generating the styling. When I need new properties, I will add them to the variables.scss file.

src/main/webapp/app/shared/components/filter/filter.component.ts
paul-luchian: I wouldn't create a custom form control for the filters, because in this case, beside the filters, I don't have anything else to add in the form. I also wanted to encapsulate the logic for setting/ reading query params in the filters component. What I wanted was to have access, from the parent, on the valueChanges event. I was too lazy to do the unsubscribe manually in the app.component ... , and that is why I went on that direction and used switchMap for canceling the old apis in que. But yes, it would had been better to expose only the valueChanges events, and to unsubscribe previous apis calls manually.

src/main/webapp/app/shared/functions/query-params.utils.ts
paul-luchian: First, I created that function for the local-task.service.ts, so I could replicate as much as possible an api call (with params, building, creating, etc), and that is why I remained set on that :)). No reason not to use the httpClient arguments.

src/main/webapp/app/shared/components/filter/filter.component.ts
paul-luchian: Changed the logic in the last commit. Only exposed: filterChanged event from the filter.component.ts

src/main/webapp/app/shared/functions/query-params.utils.ts
paul-luchian: Deleted this unused utility, and integrated httpClient -> options -> params: HttpParams. Changes present in the last commit

